PR #11, opened 2017-06-26
Proposal for a l3utils module

cereda:
Friends,

Here's a proposal of a `l3utils` module. Plan is to build a set of helper functions for general `l3` usage, as a means to ease the maintenance of the main code. For now, only cosmetic functions were added. Now, my plan is to tackle command line arguments. 😉 

The new file can also be merged into the main code. The idea behind the proposed code is to encapsulate functions inside a namespace, but this can also happen inside the other file as well without side effects.

I also tried to document the Lua functions as much as possible. Lua documentation can be generated with [LDoc](https://github.com/stevedonovan/LDoc), but I do believe the code is self-explanatory.

Have fun! 😄

### DIFF
--- a/l3utils.lua
+++ b/l3utils.lua
@@ -22,6 +22,18 @@ for those people who are interested.
 
 --]]
 
+--[[
+
+Documentation can be generated with:
+$ ldoc -f markdown l3utils.lua
+
+--]]
+
+--- Provides utility functions for the `l3build` testing and building
+--- system.
+-- The `l3utils` module provides utility functions for `l3build`. Note
+-- that functions are available through a namespace.
+
 -- the LaTeX3 utils namespace
 local l3utils = {}
 
@@ -32,7 +44,7 @@ local l3utils = {}
 -- @param value the variable value to be checked.
 -- @param default the default value to be returned in case the variable does
 -- not hold any value.
--- @return the existing value variable if it holds any value or a predefined
+-- @return The existing value variable if it holds any value or a predefined
 -- value.
 function l3utils.ensure(value, default)
   return value or default
@@ -40,9 +52,9 @@ end
 
 --- Checks if the script is running on a Windows machine.
 -- This function checks if Windows is the underlying operating system
--- by inspecting the path separator. The occurrence of '\' indicates a
+-- by inspecting the path separator. The occurrence of `\` indicates a
 -- Windows machine.
--- @return a logic value indicating whether the script is running on a
+-- @return A logic value indicating whether the script is running on a
 -- Windows machine.
 function l3utils.windows()
   return package.config:sub(1, 1) == '\\'
@@ -61,7 +73,7 @@ end
 -- reset key to restore the defaults.
 -- @param force a logic flag indicating if the colour scheme should be
 -- returned regardless of the underlying operational system.
--- @return a string containing the colour scheme for Unix-like
+-- @return A string containing the colour scheme for Unix-like
 -- terminals, or an empty string in case of Windows (can be overriden).
 function l3utils.colour(key, force)
   force = l3utils.ensure(force, false)
@@ -97,7 +109,7 @@ end
 -- @param text the string to be enclosed.
 -- @param force a logic value to force the colour scheme regardless of the
 -- underlying operating system.
--- @return the string enclosed in a colour scheme if the operating system is
+-- @return The string enclosed in a colour scheme if the operating system is
 -- not Windows, or a non-coloured output otherwise (can be overriden).
 function l3utils.coloured(key, text, force)
   force = l3utils.ensure(force, false)
@@ -109,7 +121,7 @@ end
 -- This function gets the linebreak symbol based on the underlying
 -- operating system. It was written to be as much platform-independent as
 -- possible.
--- @return the linebreak symbol (potentially '\n').
+-- @return The linebreak symbol (potentially `\n`).
 function l3utils.linebreak()
   return package.config:sub(2, 2)
 end
@@ -117,7 +129,7 @@ end
 --- Wraps a string into a sequence of lines according to a specified
 --- width.
 -- This function takes a string and splits it into a sequence of lines,
--- separated by the default linebreak symbol (potentially '\n'). Lines
+-- separated by the default linebreak symbol (potentially `\n`). Lines
 -- are broken at spaces. The logic behind this function aims at handling
 -- coloured sentences as well, but it was not tested enough. Note that
 -- Lua has some issues handling Unicode strings, so this function is
@@ -126,7 +138,7 @@ end
 -- @param width a nonzero, positive integer representing the number of
 -- colums to be displayed (in general, a sensible value would be lower
 -- than 80 columns).
--- @return the wrapped string.
+-- @return The wrapped string.
 function l3utils.wrap(text, width)
   local wrapped, colour, lb = '', '', l3utils.linebreak()
   local checkpoint, counter = 1, 1
--- a/l3utils.lua
+++ b/l3utils.lua
@@ -183,5 +183,151 @@ function l3utils.wrap(text, width)
   return wrapped
 end
 
+--- Resolves an option key based on the CLI configuration table to a
+--- normalized reference.
+-- This function takes an option key in its short or long form and
+-- returns the long form. If the option key does not exist in the CLI
+-- configuration table, a nil value is returned.
+-- @param key The option key without the dash prefix.
+-- @param configuration The CLI configuration table.
+-- @return The long form of the provided option key as a normalized
+-- reference, or a nil value if the key does not exist in the CLI
+-- configuration table.
+function l3utils.keyresolve(key, configuration)
+  for _, v in ipairs(configuration) do
+    if v['short'] == key or v['long'] == key then
+      return v['long']
+    end
+  end
+  return nil
+end
+
+--- Checks if the provided option key has arguments and, if any, the
+--- minimum and maximum values defined in the CLI configuration table.
+-- This function takes an option key and checks if such option is
+-- expected to take arguments. If so, the minimum an maximum values
+-- are returned, or a nil value otherwise. Values are set in the CLI
+-- configuration table.
+-- @param key The option key without the dash prefix.
+-- @param configuration The CLI configuration table.
+-- @return The minimum an maximum values in case the option is expected
+-- to take arguments, or a nil value otherwise, based on the CLI
+-- configuration table.
+function l3utils.parametercheck(key, configuration)
+  for _, v in ipairs(configuration) do
+    if v['long'] == key then
+      if v['parameters'] then
+        return l3utils.ensure(v['parameters']['min'], 1),
+               l3utils.ensure(v['parameters']['max'], 1)
+      else
+        return nil
+      end
+    end
+  end
+end
+
+--- Checks if the option key can take more arguments, if any.
+-- This function takes an option key, the CLI configuration table and
+-- the current list of elements, and checks if the option can take more
+-- elements. If so, the key is kept unchanged and it is returned as is,
+-- or an `unpaired` value is returned.
+-- @param key The option key.
+-- @param configuration The CLI configuration table.
+-- @param list The current list of elements referring to the provided
+-- option key.
+-- @return The unchanged key if the current list of elements can take
+-- another element, or the `unpaired` key, indicating that the exceeding
+-- arguments must be in the generic list.
+function l3utils.keycheck(key, configuration, list)
+  if key == 'unpaired' then
+    return key
+  else
+      local min, max = l3utils.parametercheck(key, configuration)
+    if min then
+      if #list < max then
+        return key
+      else
+         return 'unpaired'
+      end
+    else
+      return 'unpaired'
+    end
+  end
+end
+
+--- Classifies a list containing the actual command line arguments against
+--- a CLI configuration table.
+-- This function takes a list of elements representing the actual command
+-- line arguments and classifies it according to a CLI configuration
+-- table.
+-- @param List containing the arguments.
+-- @param configuration The CLI table configuration.
+-- @return Table containing the options and their corresponding
+-- arguments, if any. The `unpaired` option represents arguments without
+  -- associated options.
+-- @return Table containing the unknown flags.
+function l3utils.classify(arguments, configuration)
+  local result, key, err = { unpaired = {} }, 'unpaired', {}
+  local a, b
+  for _, argument in ipairs(arguments) do
+
+    a, _, b = string.find(argument, '^%-%-(%w.*)$')
+
+    if a then
+      key = l3utils.keyresolve(b, configuration)
+
+      if not key then
+        table.insert(err, b)
+        key = 'unpaired'
+      end
+
+      result[key] = l3utils.ensure(result[key], {})
+      key = l3utils.keycheck(key, configuration, result[key])
+    else
+      a, _, b = string.find(argument, '^%-(%w.*)$')
+
+      if a then
+        key = l3utils.keyresolve(b, configuration)
+        if not key then
+          table.insert(err, b)
+          key = 'unpaired'
+        end
+        result[key] = l3utils.ensure(result[key], {})
+        key = l3utils.keycheck(key, configuration, result[key])
+      else
+        key = l3utils.keycheck(key, configuration, result[key])
+        table.insert(result[key], argument)
+      end
+
+    end
+
+  end
+  return result, err
+end
+
+--- Provides a friendly representation of table elements.
+-- This function provides a pretty printing feature for representing
+-- table elements. If a value different other than `table` is provided,
+-- its string representation is returned instead.
+-- @param t The element to be printed, potentially a table.
+-- @return A textual representation of the element.
+function l3utils.prettyprint(t)
+  if type(t) ~= 'table' then
+    return tostring(t)
+  end
+
+  local f = t[1] and ipairs or pairs
+  local result = '{'
+  local comma = ''
+
+  for k, v in f(t) do
+    result = result .. comma .. ' ' .. tostring(k) ..
+             ' --> ' .. l3utils.prettyprint(v)
+    comma = ','
+  end
+
+  return result .. ' }'
+end
+
 -- export module
 return l3utils
--- a/l3utils.lua
+++ b/l3utils.lua
@@ -45,7 +45,7 @@ end
 -- @return a logic value indicating whether the script is running on a
 -- Windows machine.
 function l3utils.windows()
-  return package.config:sub(1,1) == '\\'
+  return package.config:sub(1, 1) == '\\'
 end
 
 --- Calculates the colour scheme for Unix-like terminals.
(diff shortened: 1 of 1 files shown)


l3utils.lua
```diff
@@ -0,0 +1,79 @@
...
+local l3utils = {}
+
+--- Checks if the script is running on a Windows machine.
+-- This function checks if Windows is the underlying operating system
+-- by inspecting the path separator. The occurrence of '\' indicates a
+-- Windows machine.
+-- @return a logic value indicating whether the script is running on a
+-- Windows machine.
+function l3utils.windows()
+  return package.config:sub(1,1) == '\\'
```
josephwright: In `texlua` (at least) we can test `os.type == "windows"` rather than needing to look at `package.config`.

cereda: Thanks for the pointer, I was unaware of that feature in `texlua`. I am testing my code in both interpreters, `lua` (5.3.4) and `texlua`, maybe I should stick only with the latter.

josephwright: I doubt `l3build` works with vanilla Lua: we need the `kpse` library in various places.

cereda:
Understood. It makes sense and it is a very sensible approach.

josephwright:
Perhaps I should use these comments to make a `CONTRIBUTING.md` file :)

FrankMittelbach:
Am 26.06.17 um 14:51 schrieb Joseph Wright:
> Perhaps I should use these comments to make a |CONTRIBUTING.md| file :)

well collecting that inside generally might be a good idea eventually :-)

cereda:
Hello friends. 😉 

It's been an eye opening experience, as it is my first serious Lua coding in ages. I will take a look at the `.dtx/.tex` files as well for the documentation part (I must confess it's rather unusual for me to keep actual `.lua` code comments to a minimum, but I guess I can live with that). I will postpone merging `l3utils` and `l3build` until I have a working argument parser, so I can start doing the proper refactoring. 😉

josephwright:
@cereda There's no need to remove comments in the Lua code, it's just that the 'official' documentation needs to be something that works in the TeX sphere.

cereda:
@josephwright Understood! 🍷

josephwright:
Depending on where you are 'going' with the code, it might work as a separate Lua file. I'm not 100% sure of the plan, but it looks like 'pretty printing'. We could have that as a 'try to load and fall back' option ...

cereda:
So far, the code is for pretty printing indeed, but the follow-up plan includes the addition of argument parsing. Let us wait a bit more. 😉

josephwright:
@cereda  Will await with interest. The current arg parsing stuff is somewhat inspired by a 'stand alone' approach in TeX Live, though the current code is mainly mine. I keep wondering if we should allow switches after the file list ...

cereda:
Test failed, not my fault! 😄

josephwright:
Interesting approach to dealing with arguments, but I have a few queries (and you point out one new issue I'd not spotted with the current approach, plus reminding of a second one!).

The first query I've got is whether it would be better if all of the support functions were local to the main routine: I'm not sure of the use case for them outside. The second thing that I don't currently see is how you actually split the command input: see below. Third, and not a issue just a comment: I'm interested that you don't go for a table indexed by either long or short name: possibly a good thing as I have to reverse that to output options back for recycling.

Broadly, the plan in `l3build` is to follow https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html and https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html, though at present `--version` is a bit tricky (I need a better name for the long name for `-v`). At present the parser in `l3build` has a couple of issues

- Doesn't correctly deal with `-qH` = `-q -H` (this one is new: I'd not noticed it before)
- Doesn't allow completely arbitrary ordering of arguments as files have to come last
  (not a big issue thought perhaps should be addressed)

You obviously deal with the second one nicely (requiring just taking `unpaired` and renaming as `files`), but not the first. Moreover, I'm not really sure about the 'arguments' business. The normal spec is that each option can take only one argument, which may be a comma list. Depending on the use case, that may then itself be either split at commas, which in Lua is sensible as a table, or may be a simple string (though one might say that splitting at commas should always happen). That's why in `l3build` at present I've gone for `type`: `boolean` (no arguments), `string` (take an argument, don't split) or `table` (take an argument, do split).  Using `parameters` with a max/min seems to not follow the GNU guidance, and moreover I'm not sure what happens with
```lua
local config = { short = "f", long = "foo", parameters = {min = 1, max = 2}
...
local args = {"--foo", "first", "filename", "another-filename"}
```
as the status of the `args[3]` seems to be unclear!

This all takes me back to the nature of the input at the command line, where we might currently have something like
```
texlua build.lua check --engine=pdftex,luatex -H -q m3expl001 m3expl003
```
(that `=` is optional: could also be a space!). You can of course split `args` at spaces to get some of the info, but you need to deal with (optional) `=` and `,` as well as the fact that `arg[1]` needs special handling ...

josephwright:
Note that overall your approach is probably clearer than mine, it's just it needs to do the same thing ...

josephwright:
Probably worth noting that the syntax idea of `target` (which isn't really an option but anyway) is closely modelled on Git.

josephwright:
Actually, GNU say 'Options typically precede other non-option arguments': although occasionally something like `check m3expl001 -epdftex` might be nice, I think on balance it's not vital to allow options after names.

josephwright:
I wonder if we _should_ provide `argparse()` (or similar) as a stand-alone file, perhaps `l3argparse`, with some flag to deal with a 'target' as the first argument? That I can see being of wider use (perhaps not limited to the team or even TeX users).

josephwright:
Would come out as something like
```
function argparse(args, options, [intersperse], [target])
```
where `args` and `options` are tables and `intersperse` and `target` are booleans (allow non-options and options to be mixed, and treat `arg[1]` as a 'target', respectively).

cereda:
Joseph, thanks for the feedback, it is much appreciated, as always. I am in a hurry at the moment, but I will try to address a few points you've raised throughout my new commit.

> The first query I've got is whether it would be better if all of the support functions were local to the main routine: I'm not sure of the use case for them outside.

That's a fair point. I usually favour broader, general purpose functions when writing code, but that's of course stylistic. I am beginning to suspect this utility module might be useful beyond the L3 scope, e.g, in a future update of my `checkcites` script and possibly more (ongoing or planned) Lua projects under the TeX umbrella. Again, this is just a design choice. 😉 

> The second thing that I don't currently see is how you actually split the command input: see below.

More on that later.

> Third, and not a issue just a comment: I'm interested that you don't go for a table indexed by either long or short name: possibly a good thing as I have to reverse that to output options back for recycling.

I think it's important to normalise the table keys, as you reduce the code and potentially the lookup as well. It is inevitable to find a scenario where long and short options are used, so I think we need to have a mapping to avoid key redundancy when mapping things of the same nature.

> Broadly, the plan in `l3build` is to follow https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html and https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html, though at present `--version` is a bit tricky (I need a better name for the long name for `-v`).

Command line interfaces (also known as CLI) are a very tricky and sensitive subject. To be honest, I always favour a compromise solution, as it is easier (and less complex) to cover just a subset of features than trying to provide a 1:1 specification coverage. If we are aiming for a specific scenario (in our case, `l3build`), I believe it's better for us to pinpoint what our parser identifies and how it works rather than doing the other way around, i.e, taking the command line arguments in an arbitrary order and trying to guess what to expect from it.

> You obviously deal with the second one nicely (requiring just taking `unpaired` and renaming as `files`), but not the first.

I went for `unpaired` in the sense of, well, not having an associated pair. 😉 But you are right, this list (well, table, but we get the general idea) refers to the list of files to be processed.

The first issue is easily covered by inspecting the second return value of `classify`, which handles unkown keys. We will have, e.g, `{ "qH" }`, then it is a matter to iterate over each letter and check it against the short names in our CLI configuration table. However, this form only works for switches, and to be honest I'd rather have `-q -H` than `-qH`, and our potential users should be discouraged to do such atrocity! 😉 

> Moreover, I'm not really sure about the 'arguments' business. The normal spec is that each option can take only one argument, which may be a comma list.

I think this is basically a design choice.

> Depending on the use case, that may then itself be either split at commas, which in Lua is sensible as a table, or may be a simple string (though one might say that splitting at commas should always happen).

Personally, I think this point is related to post processing, actually. At CLI parsing, a comma separated list such as `a,b,c,d` is still a single string and should be handled as such. Later on, we could apply (some would say *map*) the values into the correct references.

> That's why in `l3build` at present I've gone for `type`: `boolean` (no arguments), `string` (take an argument, don't split) or `table` (take an argument, do split).

I have a similar idea, but at the moment I wrote a classifier, not a proper parser yet. 😉 I thought of adding a post processing function, as Lua handles it as a first class element. It is almost the very same approach you took, but it is not limited to three possibilities. Found a function, apply it over each element of list.

> Using `parameters` with a max/min seems to not follow the GNU guidance, and moreover I'm not sure what happens with
>
```lua
local config = { short = "f", long = "foo", parameters = {min = 1, max = 2}
...
local args = {"--foo", "first", "filename", "another-filename"}
```
> as the status of the `args[3]` seems to be unclear!

It is clear, actually. 😉  Based on the configuration, `--foo` is expected to get at most 2 elements (`max` is used for this, `min` has no influence here, it would have in a post processing, e.g parsing). In our case, `args[3]` would be put inside `unpaired`.

Regarding the GNU guidance, this is certainly not standard, but I must say it helps a lot. In general, it will suffice, for most of our stuff, to write

```lua
local config = { short = "b", long = "bar" = parameters = {} }
```

which is equivalent to

```lua
local config = { short = "b", long = "bar" = parameters = { min = 1, max = 1} }
```

Accepting more than one parameter is just a syntactic sugar for CLI (if that exists, of course). My plan was to provide a general way of reducing post processing when it could be done as the syntactic analysis is already in progress. Say

```lua
local config = { short = "r", long = "resize" = parameters = { min = 2, max = 2} }
```

allows us to write

```
$ video.lua --resize 800 600 input.avi
```

instead of

```
$ video.lua --resize -w 800 -h 600 input.avi
```

Again, it's stylistic, not a blocker. 😉  Perhaps it will save us some processing later on, but I have no idea right now!

> This all takes me back to the nature of the input at the command line, where we might currently have something like
```
texlua build.lua check --engine=pdftex,luatex -H -q m3expl001 m3expl003
```
> (that `=` is optional: could also be a space!). You can of course split `args` at spaces to get some of the info, but you need to deal with (optional) `=` and `,` as well as the fact that `arg[1]` needs special handling...

Ah the separators. I thought of them, but decided not to mention the possibility as it would significantly increase my work. 😉  Certainly doable, a `separator = '='` is in our plans, and I like the challenge! That said, what would be the practical difference of

```
texlua build.lua check --engine=pdftex,luatex -H -q m3expl001 m3expl003
```

to

```
texlua build.lua check --engine pdftex,luatex -H -q m3expl001 m3expl003
```

or even

```
texlua build.lua check --engine pdftex --engine luatex -H -q m3expl001 m3expl003
```

and similar approaches? Note that the current code already catches multiple key entries, so

```lua
engine = { 'pdflatex', 'luatex' }
```

will be there without any additional effort. Of course, there's the `min/max` discussion (we could use a high value for `max`), which brings me to the very question mentioned earlier: should we go 1:1 standard or adopt a subset of features which are "good enough"? I think it's important to tell the rules to the users, so everything beyond good usage is their fault, not ours. David will surely like this last bit! 😉

josephwright:
>> The first query I've got is whether it would be better if all of the support functions were local to the main routine: I'm not sure of the use case for them outside.
> 
> That's a fair point. I usually favour broader, general purpose functions when writing code, but that's of course stylistic. I am beginning to suspect this utility module might be useful beyond the L3 scope, e.g, in a future update of my checkcites script and possibly more (ongoing or planned) Lua projects under the TeX umbrella. Again, this is just a design choice. 😉

I also started with 'generally global' as that's what I'm used to from TeX. However, it means you have to assume that everything might get used by others so needs documenting and a fixed API. For more internal stuff I'm not sure I'd go for that.

On 'utilities', I think that might be best dealt with separately:  back with the point that `l3build` is deliberately very self-contained. I'm not sure of exactly what you want, but for L3 we do have some minimal Lua and could look to extend there if there is a 'TeX link'. For 'pure' scripting, perhaps a new product?

>> Third, and not a issue just a comment: I'm interested that you don't go for a table indexed by either long or short name: possibly a good thing as I have to reverse that to output options back for recycling.
> 
> I think it's important to normalise the table keys, as you reduce the code and potentially the lookup as well. It is inevitable to find a scenario where long and short options are used, so I think we need to have a mapping to avoid key redundancy when mapping things of the same nature.

Sure: I meant thought that one can do `["long"] = {short = "s"}` which still maps the names. Not a big thing, just a question of style.

>> Broadly, the plan in l3build is to follow https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html and https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html, though at present --version is a bit tricky (I need a better name for the long name for -v).
> 
> Command line interfaces (also known as CLI) are a very tricky and sensitive subject. To be honest, I always favour a compromise solution, as it is easier (and less complex) to cover just a subset of features than trying to provide a 1:1 specification coverage. If we are aiming for a specific scenario (in our case, l3build), I believe it's better for us to pinpoint what our parser identifies and how it works rather than doing the other way around, i.e, taking the command line arguments in an arbitrary order and trying to guess what to expect from it.

You make a good case for _not_ trying to make a generalised `argparse` but rather to agree on a spec for `l3build` and keep the code there (with comments so other people can pick up the ideas and adapt).  I'll come back to that below.

>> You obviously deal with the second one nicely (requiring just taking unpaired and renaming as files), but not the first.
> 
> I went for unpaired in the sense of, well, not having an associated pair. 😉 But you are right, this list (well, table, but we get the general idea) refers to the list of files to be processed.

The name we can change if we need to ;)

> The first issue is easily covered by inspecting the second return value of classify, which handles unkown keys. We will have, e.g, `{ "qH" }`, then it is a matter to iterate over each letter and check it against the short names in our CLI configuration table. However, this form only works for switches, and to be honest I'd rather have `-q -H `than `-qH`, and our potential users should be discouraged to do such atrocity! 😉

Agreed: we can simply stick to one at a time.

>> Moreover, I'm not really sure about the 'arguments' business. The normal spec is that each option can take only one argument, which may be a comma list.
> 
> I think this is basically a design choice.
> 
>> Depending on the use case, that may then itself be either split at commas, which in Lua is sensible as a table, or may be a simple string (though one might say that splitting at commas should always happen).
> 
> Personally, I think this point is related to post processing, actually. At CLI parsing, a comma separated list such as a,b,c,d is still a single string and should be handled as such. Later on, we could apply (some would say map) the values into the correct references.

OK in conceptual terms, but for `l3build` the actual uses are clear: the things that are strings don't have comma, the things that are multi-valued are comma lists that turn into tables. I'd rather not have to re-write the rest of the code and convert in each place they are used ...

>> That's why in l3build at present I've gone for type: boolean (no arguments), string (take an argument, don't split) or table (take an argument, do split).
> 
> I have a similar idea, but at the moment I wrote a classifier, not a proper parser yet. 😉 I thought of adding a post processing function, as Lua handles it as a first class element. It is almost the very same approach you took, but it is not limited to three possibilities. Found a function, apply it over each element of list.

This I will trust you on: to me it's more natural to state the type of option 'up front' (much like keyvals).

>> Using parameters with a max/min seems to not follow the GNU guidance, and moreover I'm not sure what happens with
> 
> local config = { short = "f", long = "foo", parameters = {min = 1, max = 2}
> ...
> local args = {"--foo", "first", "filename", "another-filename"}
> 
>     as the status of the args[3] seems to be unclear!
> 
> It is clear, actually. 😉 Based on the configuration, --foo is expected to get at most 2 elements (max is used for this, min has no influence here, it would have in a post processing, e.g parsing). In our case, args[3] would be put inside unpaired.

Still don't see it: are you saying the approach is lazy, so only takes `first` but not `filename` as it doesn't _have_ to?

> Regarding the GNU guidance, this is certainly not standard, but I must say it helps a lot. In general, it will suffice, for most of our stuff, to write
> 
> local config = { short = "b", long = "bar", parameters = {} }
> 
> which is equivalent to
> 
> local config = { short = "b", long = "bar", parameters = { min = 1, max = 1 } }
> 
> Accepting more than one parameter is just a syntactic sugar for CLI (if that exists, of course). My plan was to provide a general way of reducing post processing when it could be done as the syntactic analysis is already in progress. Say
> 
> local config = { short = "r", long = "resize", parameters = { min = 2, max = 2 } }
> 
> allows us to write
> 
> $ video.lua --resize 800 600 input.avi
> 
> instead of
> 
> $ video.lua --resize -w 800 -h 600 input.avi

Back with the earlier point: for `l3build` the GNU-like (Git-based) form applies.

>> This all takes me back to the nature of the input at the command line, where we might currently have something like
>> 
>>    texlua build.lua check --engine=pdftex,luatex -H -q m3expl001 m3expl003
>> 
>>  (that = is optional: could also be a space!). You can of course split args at spaces to get some of the info, but you need to deal with (optional) = and , as well as the fact that arg[1] needs special handling...
> 
> Ah the separators. I thought of them, but decided not to mention the possibility as it would significantly increase my work. 😉 Certainly doable, a separator = '=' is in our plans, and I like the challenge! That said, what would be the practical difference of
> 
>    texlua build.lua check --engine=pdftex,luatex -H -q m3expl001 m3expl003
> 
> to
> 
>     texlua build.lua check --engine pdftex,luatex -H -q m3expl001 m3expl003

At present no as `=` is mainly useful for cases where the argument to the option is itself optional and we don't have any of those. However, it's part of the 'usual' syntax for `--`-type arguments, so needs to work.

>    texlua build.lua check --engine pdftex --engine luatex -H -q m3expl001 m3expl003

This one is currently 'undefined': I guess I favour a keyval-like view that there is only one value, so just `luatex` in the above.

That all brings me to the conclusions. I think you've made a good case for keep the argparse functionality in `l3build` not in a separate file, so we then need to decide what form any replacement for the current version takes. The syntax we have is

- `arg[1]` is either a target or an option (and for the latter only `--help|-h` and `--version` really make sense) (I'll rename again the long form of `-v` to avoid `--version`)
- Options are either short `-<letter>` or long `--<name-which-may-take-hyphens>`
- Options may take an argument, which can either run directly on from the switch or be separated by a space. For long options (only) an optional `=` may be present, e.g. the following are all valid: `-epdftex`, `-e pdftex`, `--engine pdftex`, `--engine=pdftex`
- Where the argument to an option takes multiple values, these are comma-separated, _e.g._ `-epdftex,luatex`: these cases return the argument as a table
- A repeated option is the same as giving only the _last_ one
- Short options _cannot_ be run together (_e.g._ -qH` is an error)
- `--` on its own terminates options and can thus be used for non-option arguments starting `--`
- Options come _before_ non-option arguments (file names)

josephwright:
I think here we are unlikely to use this code in `l3build` in the current form (though the ideas may well be useful in a separate place): I'll close the pull request if that's OK.

cereda:
Sure thing, Joseph. Fire at will.

PR #26, opened 2017-07-21
Do not typeset the same documents twice

dffischer:
The `doc` function searches for the `typesetfiles` and `typesetdemofiles` in both `unpackdir` and `typesetdir`. When a file shows up in both, because it was matched by both `sourcefiles` and `unpackfiles` or copied from `unpackdir` to `typesetdir`, the same file is processed twice from both locations. As the same file leads to the same PDF output, this unnecessarily wastes time running TeX and can be saved.


dffischer:

PR #28, opened 2017-07-21
Use the basename function instead of its contents

dffischer:
This just eliminates a small code duplication. Most probably it was overlooked while applying pull requests that changed code in the same vicinity.


dffischer:

PR #29, opened 2017-07-21
Accessing files in subdirectories

dffischer:
This adjusts the following two related routines to correctly deal with files in subdirectories.
- copying out typeset files from the `build` directory to the project root
- filtering which files to typeset using command-line-arguments


dffischer:
I rebased the branch to cleanly apply after the recent merges.

dffischer:

PR #30, opened 2017-07-22
resolve copied links

dffischer:
When any file copied into a build directory is a symbolic link, the link could not correctly point to its target from its new location, should the target not coincidentally be copied as well. Errors of the kind "file not found" will then occur for the copied links.  This is true for targets inside as well as outside of the project, as long as they are specified relatively.

Resolving the links, copying the file content of the target instead of the link, eliminates this problem.

I did not test this under Windows, but as far as I can tell from [the documentation of the xcopy utility](https://technet.microsoft.com/de-de/library/cc771254(v=ws.10).aspx), specifically the description of the `/b` option, this was the default behavior there all along.

### DIFF
--- a/l3build.lua
+++ b/l3build.lua
@@ -611,7 +611,7 @@ function cp(glob, source, dest)
         )
       end
     else
-      errorlevel = execute("cp -rf " .. source .. " " .. dest)
+      errorlevel = execute("cp -rLf " .. source .. " " .. dest)
     end
     if errorlevel ~=0 then
       return errorlevel


josephwright:
I see the point _but_ links are not cross-platform and they are not allowed for TeX Live (so basically not for TeX use).

wspr:
But if it fixes a problem, isn’t that a good thing? Just because symbolic links aren’t used in TeX Live doesn’t mean people won’t/can’t use them personally.

Especially if testing reveals this pull request makes the behaviour equal on Windows and *nix I’d be inclined to favour it.

dffischer:
In fact, the change eliminates links from the installed files. It allows the use of links inside the sources, and when they are copied into the build environment, they are transformed (resolved) into independent files.

An exemplary use case would be a project with many small subpackages that all use the same simple `*.ins` file. This file could then be kept in a single location and symlinked into all the subpackage directories.

josephwright:
I'm still slightly wary but probably it won't do any active harm, so I'll merge.

dffischer:
Thank you for the merge.

FrankMittelbach:
Am 23.07.17 um 12:32 schrieb Joseph Wright:
> I'm still slightly wary but probably it won't do any active harm, so 
> I'll merge.

it is probably a Good Thing (TM) but I would feel more comfortable if 
such a behavior is actually documented as part of the doc section on how 
the system behaves

josephwright:
OK: I'll make a note of that in the docs later today.

PR #32, opened 2017-09-11
Allow the "error level" to be recorded in TLG files

wspr:
When testing a package, it's good to be able to check whether a TeX error has been tripped even if it's not caught inside \TIMO..\OMIT. This code keeps track of the error levels from each typesetting run and simply appends them to the TLG file if a conditional ("recorderrorlevel") is set true.

It might also make sense to include the status from bibtex compilations etc but I needed to keep things simple for the time being... 

Happy to discuss naming. Not sure if "error level" is canonical. Could also use "exit status". I think getting the text right here (that's appended to the TLG) is also important before it's rolled out, and not sure if what I've written is best...

### DIFF
--- a/l3build.dtx
+++ b/l3build.dtx
@@ -148,6 +148,7 @@
 \luavarset{scriptname}  {"build.lua"} {Name of script used in dependencies.}
 \luavarset{typesetcmds} {""}          {Instructions to be passed to \TeX{} when doing typesetting.}
 \luavarset{versionform} {""}          {Nature of version strings for auto-replacement.}
+\luavarset{recorderrorlevel} {false}  {Include error level(s) from test run(s) in TLG files?}
 }
 \allluavars
 \newcommand\luavartypeset{%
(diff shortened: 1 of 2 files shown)


josephwright:
I might try to come up with a slightly shorter name, but otherwise looks fine. I'll merge a bit later on today.

FrankMittelbach:
> Happy to discuss naming. Not sure if "error level" is canonical. Could 
> also use "exit status". I think getting the text right here (that's 
> appended to the TLG) is also important before it's rolled out, and not 
> sure if what I've written is best...

call it exit status or something  similar (as it may not be an error) 
and that should also be reflected in

+        write('Run ' .. i .. ' executed without error')
+      else
+        write('Run ' .. i .. ' executed with error level ' .. 
errlevels[i] )

which then may not even need a distinction in the wording

josephwright:
Good point: perhaps `savestatus`?

wspr:
On 11 Sep 2017, at 10:59 pm, Joseph Wright <notifications@github.com> wrote:
> 
> Good point: perhaps savestatus?

I’m happy with that, and this morning on the walk to work I came to the same conclusion as Frank.
“Compilation of test file completed with exit status X” or similar would work well I think.

Cheers,
Will

wspr:
I couldn't decide between `recordstatus` / `savestatus` — otherwise updated as per the above.

wspr:
Thanks. Hope you don’t mind the approach of pull request–merge … I think it makes for a nice approach to discuss new features / perform code review. Not saying we need to do it for all development work though :)

josephwright:
@wspr  It's probably more of the 'Git model' :) As you have access to the main repo you could have made a branch here and still made a pull request - doesn't really matter as we don't have a formal policy one way or the other. (You see bigger projects where all code has to be added _via_ pull requests: overkill for us I suspect.)

Certainly for stuff where a more formal review is sensible it works well: one of the reasons I'm broadly convinced of Git (or similar) working well compared with say SVN.

wspr:
Oh, I (foolishly) hadn’t realised you could pull request from within the same repository. That would make far more sense!

josephwright:
@wspr  Not a problem :) Like I said, some projects got very much with 'the main repo is only touched to bring in reviewed code', so it's a valid working model. We probably don't need anything so complex :)

PR #34, opened 2017-11-09
Add "manifest" target

wspr:
## Status

Under development (not yet for rebasing)

## Description

I thought it would be useful for l3build to be able to produce a "manifest" file for more complex packages.

For example, rather than trying to manually enumerate all files for the copyright/license text, it would make to generate a file with l3build based on the rules we know l3build will be using to process the package.

## Todo

- [x] Provide interface for customisation
- [x] Ensure all user-level features and customisations are documented


josephwright:
At a technical level a quick look suggests all is OK (I'll look carefully later): what I wonder is whether anyone uses these manifests practically.

josephwright:
I guess I'd like to see the mappings (text <-> variables) be themselves part of the adjustable set up, and so documented.

josephwright:
Ah: have you considered generated files? For example, testing on `l3build` itself the PDF doesn't show up!

wspr:
## Regarding "practicality"

I see two selling points (only, unfortunately):

1. Rigour for the copyright and license statements ("this package consists of all files listed in MANIFEST", for example)

1. Debugging (and checking) what l3build is doing — e.g., after updating `build.lua` I might start dropping a file or two, and if this is picked up by the manifest, then it saves a headache for the CTAN people.

## On generated files

This one is a lot harder! I guess it requires a full `build ctan` so that all generated files are picked up correctly. That's probably a good idea, really.

## On mappings

I agree this needs to be less ad-hoc, but at the same time I think it should be governed by the l3build logic itself.

josephwright:
I wonder if the MANIFEST file should be created only in the build tree? Are you seeing it as 'dynamic' (so never checked in) or more akin to the `setversion` target (editing files that can be then checked in)?

wspr:
I was largely thinking it would be more like the latter — I wanted a way to avoid having hard-coded info in package files like

```
This work consists of the file  siunitx.dtx
          and the derived files siunitx.pdf,
                                siunitx.sty and
                                siunitx.ins.

```

In the past I eschewed a lot of formality around this stuff, but I think I've matured a bit in that sense — I'm now thinking it's better to have the copyright and licensing all strictly proper, to the letter, even if it's a bit over the top. (E.g., copyright statements in each file, and so on.)

The other advantage I see in the long run would be to also provide hooks for documentation purposes. So rather than just a dumb list of files it might be sorted into customisable categories and/or with a little blurb for each file.

So just checking — are okay if I keep this on the boil to be added some time down the track?

wspr:
Another example for inspiration: https://github.com/latex3/latex2e/blob/master/required/graphics/README.md

I've started to add additional options to allow "file descriptions" to be extracted from the file itself using a `string.match` function.

I think you're right the file groupings need to be sortable and customisable. I'm not sure the best way to ensure sorting order, etc., at the moment! Lua is a bit strange with how its tables are unsorted :)

wspr:
And a longer-form example of the current functionality (of course to be improved): https://github.com/wspr/unicode-math/blob/working/MANIFEST.md

josephwright:
I've been thinking about this in terms of the interface. I like the idea of a `manifest` target but I'm not keen on the complex variables needed at present to set it up. I wonder if it would be better to use the same approach as for `setversion`: document a 'swappable' `manifest()` function:
```lua
manifest = manifest or function ()
...
```
so that users can do whatever they want to alter the output, including working with complex file situations, and we can have a relatively 'clean' set up (a simple `manifestfile` setting for the file name itself).

FrankMittelbach:
sounds like a good approach to me to provide a usable standard up front with an overwrite for complex situations

wspr:
No complaints from me. Some of the options I've added are probably a bit over the top, but I wanted to test things out before finalising an interface. In that sense I should probably worry a lot less about documentation until we know what's worth keeping…

wspr:
So before I start trying to strip back the current setup, the main concern I have now is around customisation like

```
manifestgroups = manifestgroups or {
-- this needs to be an array of tables, not a table of tables, to ensure ordering.
    {
       name    = "Source files",
       description = [[
These are source files for a number of purposes, including the `unpack`
process which generates the installation files of the package. Additional
files included here will also be installed for processing such as testing.
       ]],
       files   = {sourcefiles},
    },
```

If you want to take this out into your own `build.lua` file to setup some custom groups, `sourcefiles` and other l3build variables haven't yet been defined, so all file patterns would need to be input "explicitly" as `files   = {"*.dtx","*.ins"}` and so on. Do you think that's okay for the time being?

josephwright:
@wspr  Like I said, I'd probably not provide a range of options at all, just a 'hard coded' function for the generation. You can then say 'See the function `XXX` , which may be customised by definition an alternative in the `build.lua` file'. It can then use `sourcefiles`, _etc._ directly.

wspr:
I agree we want to keep things simple. You still need to get the data into a table format through. So for a function-based approach to work you’d have function(s) that take a long list of input arguments for all the `xyzfiles` and `xyzdir` variables? Basically something like:

```
buildmanifestgroups = buildmanifestgroups or function(sourcesfiles,typesetfiles,docfiles,…(ad nauseam),…,maindir,supportdir,...)

manifestgroups = {
    {
       name    = "Source files",
       description = [[...]], -- if we use this field at all...
       files   = {sourcefiles},
    },
    …
}

return manifestgroups

end
```

josephwright:
@wspr  I was thinking much more simple than that! I am imaging a single user variable
```lua
manifestfile = manifestfile or "MANIFEST.md"
```
Then having
```lua
manifest = manifest or function()
-- All of the construction here
-- including the decisions about which sources to list
-- fixed text for the manifest, etc.
end
```
_i.e._ an entirely 'fixed' form of manifest 'out of the box', where the user can if they like simply re-implement the entire thing. It will only need the (documented) variable names anyway, so it is a simple task in Lua.

The advantage of this is we don't end up trying to second-guess what people would want. (There is one variable we might want: a switch to include support/test/non-release files. I'd for example not list my `.lvt` files as part of the package as they don't show up on CTAN. Others might want to ...)

wspr:
Oh… that's embarrassing! I thought I had checked whether that worked. To be honest I'm a little baffled how it does:

```
f = function() print(x) end
x=1 f()
x=2 f()
```

I would have assumed `x`, not being a function argument, would be frozen in place at time of definition. Evidently not. So that's great. I'll get to work, later this week.

wspr:
(Thanks for looking into all this and for your patience with me, Joseph!)

josephwright:
Provided `x` is defined before `f()` we are OK. Remember that in Lua
```lua
function foo(a, b)
...
end
```
is just a shortcut for
```lua
f = function (a,b) ... end
```

josephwright:
@wspr  No problem: the entire point of having some branching and review for this type of bigger change is to allow discussion with some real code to work from. I plan to do the same to address the `--config` idea.

wspr:
Sorry this stalled -- work got unexpectedly busy and my son brought home a nasty cold for us :)
I've got a pretty clear picture here of what's needed, I just had to find the time to do it!

josephwright:
@wspr  Lots of impressive stuff!

I have I guess a couple of 'big' queries. The first is how much customisation is sensible without the user having to re-write `manifest()`. My starting point was that `manifest()` would be entirely fixed, and that any customisation at all would need an alternative version to be written. Somewhat modified from that is I guess what happens with `setversion()`, where a few 'common' cases are covered in `l3build` but more complex ones have to be handled by writing custom line parsers. I guess here it's more toward the 'never have to alter `manifest()` approach, but there are a lot of things to describe as a result.

The second is I'm not clear on what the manifest is exactly of. There are various ways one could look at this

- What goes to CTAN (source files + 'final' PDFs)
- What goes into a TeX system (source files + 'final' PDFs + DocStrip output)
- What is in some source repo (source files + test files + support files)

We probably need to be clear on which of these we are aiming at: I wasn't 100% sure of the plan at present as both PDFs (including 'transitory' demo ones) and test files are mentioned.

wspr:
@josephwright Thanks for taking a look...

I definitely hear (and heard) you about having a fixed `manifest()` function at the top level, and there's possibly still more I could do to simplify things.

Although I didn't try to prototype it, the reason I didn't take it back to that was that even without some of the fancier things I've added the plain `manifest` function had to do a lot of (tedious) work to handle the different cases we have in l3build. So my guy feeling (and probably I should have done something a bit more concrete here) was that it wasn't very feasible for most people to re-write `manifest()` entirely just to do some customisations. (And it wouldn't be very elegant or maintainable to copy/paste the entire `manifest()` function locally just to change a few strings here and there.)

So it seemed better on balance to start factoring most of the functionality into small pieces which could be swapped out as necessary.

Something else I considered was whether an array of tables was the right way to set up the "groups", that is

```
manifest_setup = manifest_setup or function()
  local groups = {
    {
       name    = "Source files",
       description = [[ ... ]],
       files   = {sourcefiles},
    },
    ...
```

A potential alternative could be to do this as a function more like

```
manifest_process_group("Source files",{sourcefiles},...)
manifest_process_group("Text files",{textfiles},...)
...
```

but while this would avoid a loop it didn't seem as idiomatic.

So I got reached a point where I was thinking that to have a fixed `manifest()` function it would need to have an extremely simplified structure like:

```
manifest = function()

<set up table of groups with their names and search patterns>
<loop over the table constructing the file lists>
<perform the write IO>

end
```

so at the very least it made sense to split up the table with the rest, and then it seemed to still make sense to not hard-code everything directly.

### What should go into a manifest?

At the moment I've erred on the side of being comprehensive. So to answer your question, perhaps a superset of your possibilities. Listing test files might seem useless, but if you add a description line for each then it becomes a way to summarise the list of tests in an easy way. (Currently the tests are pretty opaque when you take a look at, say, the 2e sources.)

* * *

I just had someone walk into my office so I'll have to leave it there... I'll try and revise later today...

wspr:
Would you like me to make another branch based on what we've got with an entirely stripped down version without any optional features to see what a truly minimal `manifest()` would look like?

josephwright:
@wspr  Thinking about the task, it's a lot more involved than `setversion()` so I can see a strong argument for wanting to avoid copy-pasting code. That said, I'm happy to map out a very 'minimal' version of the same idea so we can see what makes most sense.

wspr:
Okay, here's a largely minimal version of the current state of the manifest code:

  <https://github.com/latex3/l3build/blob/manifest-minimal/l3build-manifest.lua>

The only thing that are realistically customisable are: the groups based on a setup function and a monolithic (but only 20 lines) function that does the formatting + writing to file.

I've also taken out descriptions of both groups and files; all up this has roughly halved the length of the code. Comparing to the current state here, what looks better to you?

  <  <https://github.com/latex3/l3build/blob/manifest/l3build-manifest.lua>

wspr:
( Noting that I'm rather partial to the functionality provided by "file descriptions" as highlighted in the following example: <https://github.com/wspr/unicode-math/blob/working/MANIFEST.md> . )

josephwright:
The simper version is nice but on balance I think if we are doing this we may as well have the entire thing, in particular if we are going to split into multiple files. (The parallel suggestion for that looks good.)

One thing that occurs to me links to the 'what is the key aim': do we need two types of manifest? There is the 'files that get distributed' view and the 'files in the source repo' view (the latter probably being the place for descriptions). One might imagine that some users would like both ...

josephwright:
One thing we might want is at least a couple of 'out of the box' possibilities for the type of manifest. For example, the two 'limiting' cases of 'the files that go to CTAN' and 'the files in the repo'. That could be in the docs or could be a couple of pre-defined tables one can select from ...

josephwright:
I'm reading over the 'full fat' code in detail. I have a few queries, some of which I have a feeling reflect the fact I tend to want Lua to work like TeX ;)

I'd like to focus on one here as it's a bit involved: mapping between file lists and directories. In the rest of `l3build`, the particular file lists to use for each operation are coded directly into the respective functions, so the matching directories are simply coded in too.  ('Simply' is relative: the recent work on `sourcefiledir` was 'interesting'.)

In the set up table for manifests, files are grouped by description, so can have multiple file lists. Making sure that these always search the same places as the rest of `l3build` needs a little care I suspect. (I note that there is an optional `dir` entry, but I'm not sure that's quite what I'm getting at. In particular, I'm mindful that any assumptions about mappings should carry though the whole set up.)

It seems to me that if one wants to allow a 'fully flexible' manifest set up then we need to make sure that the code itself can link `sourcefiles` to `sourcefiledir`, _etc._ That is all doable but depending one exactly what we want looks 'interesting. Using the global `_G` table one can arrange to convert strings to variables, and metatables can be used to get the name used to call a table. Between those it should be possible to construct something like
```lua
dirmappings =
  {
    sourcefile = sourcfiledir, -- or perhaps "sourcefildir" as a string
   ...
  }
```
and to do the lookup. Of course, once one does that it becomes possible (though perhaps painful) to allow flexible mappings everywhere else ... at the moment that's not what I'm suggesting?

How do you see this: am I over-thinking it?

wspr:
@josephwright — I had similar thoughts that it might have been a good idea to link file variables with directories. Since I was (and still am) unfamiliar with metatables, my only thought at the time was that we'd need to replace the string variables with something more like

    sourcefiles = { match = {"*.dtx"}, dir = {code/} }

and that was too big a change to consider suggesting at the time! 

I do like the idea of having a mapping like you suggest. This would (mostly) avoid the need to have explicit `dir` settings in the manifest setup. OTOH, it's not too bad having the setup duplicated in the manifest mappings, so for the time being I'd be happy to leave things as they are and worry about a grand unified variable approach as a medium term change.

What do you think?

(BTW it's still on my internal todo list to update the current manifest code to include `sourcefiledir` and so on.)

wspr:
@josephwright — A couple more things to consider about unifying how the matching variables "work":

* Currently the way PDF files are specified is somewhat non-standard (matching the typeset target and then swapping extensions over). I'd consider changing this to "*.pdf" in the `build/doc` directory. This would remove the need for the `rename` setting in the manifest code (simplify and save a few lines).

* Matching test files and outputs is also done differently than the rest — they could also change to something like `*.tlg` and so on. Changing this wouldn't have a major impact on the manifest code.

* Having a `testsuppdir` match everything inside is different than the rest of the ‘support’ variable setups, although it does make things easy. (And maybe `testsuppdir` would be better as `checksuppdir` ? We seem to have three variables starting with `test` and nine containing `check`.)

You're right that the manifest code currently CANNOT generate something like:

> ## Check support files
>
> These are files used when running the test suite.
>
> * support/foo.tex
> * support/bar.tex
> * testfiles/support/testclass.cls

This could only really be done by re-thinking how the manifest setup table is currently structured.

josephwright:
@wspr Various answers :)

- I think have `sourcefildir` as a simple table of globs is about right: the idea of splitting up directories is rather separate from changing the file types involved. So I don't think we would have been better with something like `sourcefiles = { match = {"*.dtx"}, dir = {code/} }`.

- As long as the suggested structure  can be adapted if/when we make the file <-> dir mapping into a table, I'm happy to go with the current suggestion. I think it is doable ...

- On the `check`/`test` business: I know it's a bit odd! There's a sort-of logic that 'test' refers to individual `.tlg` files whereas 'check' is stuff that can cut across. I agree though that it's a bit of a mess. We could/should clean up: I'll open a separate issue for that.

josephwright:
@wspr  OK, next set of questions :)

- Is there a reason for having `manifest_setup` at all? Couldn't you just have `manifest_list = manifest_list or { ... }` is the same way we do for the existing configuration variables?

- On `manifest_list`, I guess for the 'standard' config variables I'd favour a 'simple name, and probably avoid 'list' (which is used internally where a set of file globs has to be turned into an explicit set of file names). Something like `manifestmapping`  might be appropriate ...
- I'm not sure what `manifest_build_init()` is about: you seem to be re-writing the same table with different data? Shouldn't the `foo = foo or <default>` stuff all be at point-of use rather than altering the input data? (For example, I was surprised that you didn't have `dir = manifest_list["dir"] or maindir` when you use the directory, as the `dir` value seemed to be optional.) 

- Particularly looking to the future (the directory mapping business), I'm wary of the bit of `manifest_build_init()` that says 'allow nested tables by requiring two levels of nesting'. I think that's being used to allow `files   = {"*"..lvtext,"*"..lveext,"*"..tlgext},` rather than adding some more tables to the main set up to support listing tests.  That will get tricky if/when we sort out the directory mappings, and also we are back with the logic not being the same as in the rest of `l3build`. (We should presumably have a set of tables that cover those concepts, rather than assuming all tests are live.)

- Does `extractfiledesc` need to apply per-file or indeed at all? As the user has to supply `manifest_extract_filedesc`, I'd have expected it simply to be written such that only files with the right data can be 'hits'. Counter-argument I guess is that for version setting (soon to be renamed ...) we do have `versionfiles` to limit the scope. Assuming  `extractfiledesc` is needed, wouldn't a default of `false` be sensible?

- Presumably once everything is agreed you plan to add a significant amount of documentation on this: there are a lot of functions to list ...

wspr:
I edited your comment to be an enumerated list :)

1. Yes, this goes back to the old problem of when variables are defined. If we write `manifest_list = {...}` as an ordinary table and it contains references to `sourcefiles`, the `sourcefiles` variable needs to have already been defined.

I.e., I can write

```
-- in build.lua
manifest_setup = function()
  return { {name="Source",files={sourcefiles} } }
end
...
-- in l3build.lua
sourcesfiles = "*.dtx"
manifest_list = manifest_setup()
```

but I *can't* write
```
-- in build.lua
manifest_list = { {name="Source",files={sourcefiles} }
...
-- in l3build.lua
sourcesfiles = "*.dtx"
```

1. Yes, I agree a better name would be good for this variable. I had that in the back of my mind but hadn't yet thought it through properly :)

1. Maybe my indirection here is confusing. Functionally it's exactly the same as writing `manifest_list["dir"] or maindir` at point of use, I just preferred setting up all the defaults explicitly all in one place.

1. Yes and no; I *think* I was simply writing this to be defensive, so that if a user writes `files=sourcefiles` instead of `files={sourcefiles}` it still works. Happy to drop it.

1. I've actually wondered today whether the file extraction business should be set up on a per-group basis, but then I figured you'd just write a "smarter" file extraction function. Perhaps I need to rethink this a little. It's `false` by default because (a) without an extraction function it's equivalent to being off by default already, and (b) so once there is an extraction function that you don't need to manually turn it on for every group. Probably it would be better to rename the setting to something like `skipfiledescription`.

* * *

I can't look at this tonight I'm afraid but will come back to this tomorrow or so...

wspr:
@josephwright Typo / clarification of above; the "nested table" thing means that the user can either type either:

    files = {sourcesfiles}
    files = {"*.dtx"}

It didn't seem natural to me to require that if you want to enter in manual patterns one would have to write

    files = {{"*.dtx}}

Now, maybe my way of dealing with that isn't great. In some other languages I've seen a "flatten" command that takes nested lists and strips out all the nesting. Something equivalent here would be good, but I think would take *more* code than what I've written :)

wspr:
I have also wondered whether a per-group sorting function would be better / necessary in some cases...

FrankMittelbach:
Am 13.12.2017 um 14:27 schrieb Joseph Wright:
> One thing that occurs to me links to the 'what is the key aim': do we 
> need two types of manifest? There is the 'files that get distributed' 
> view and the 'files in the source repo' view (the latter probably being 
> the place for descriptions). One might imagine that some users would 
> like both ...


haven't yet tried to actually play with that work, but, just from the 
outside, a "manifest that is on CTAN" reflecting what somebody packages 
as a distribution (even if later put into separate places in TL etc) is 
something that in the past has been used to do things like writing

This work consists of files listed in manifest ... bla bla

On the other hand having a manifest (for yourself or for anybody who 
visits a source repository) that actually lists sorts all files in that 
repository may also be nice.

however, in my view the first usage is the more important one.

FrankMittelbach:
Am 14.12.2017 um 12:41 schrieb Will Robertson:
> Yes and no; I /think/ I was simply writing this to be defensive, so that 
> if a user writes |files=sourcefiles| instead of |files={sourcefiles}| it 
> still works. Happy to drop it.

if you want to be defensive can't you in the right places check that 
something is a table by looking at the type or so?

wspr:
@FrankMittelbach — that's basically what I am doing:

```
  if type(entry.files[1])=="string" then
    entry.files = {entry.files}
  end
```

I think Joseph's point is that currently there's nothing explicit that ties together source files and source directories, so if we move to a more formalised structure we might not want arbitrary mappings in there. But I don't think at the moment we're proposing an overhaul about l3build's setup code, so I've taken it as a comment for the future.

wspr:
@FrankMittelbach — Perhaps it would make sense to have a way of designating if files are part of the CTAN bundle or not. And/or (did Joseph already suggest this?) providing two manifest files, one for the distribution and one for CTAN.

FrankMittelbach:
Am 18.12.2017 um 23:18 schrieb Will Robertson:
> @FrankMittelbach <https://github.com/frankmittelbach> — Perhaps it would 
> make sense to have a way of designating if files are part of the CTAN 
> bundle or not. And/or (did Joseph already suggest this?) providing two 
> manifest files, one for the distribution and one for CTAN.

you either want 2 files (so that you can send the CTAN one along with 
the distribution but not the other one)

or you want only one (in which case I think it should be the "release 
one not the repository one), but 2 is perhaps best.

if you want to go fancy then it might be actually very nice to indicate 
the "distribution files" with a special marker in the full manifest, so 
you have a nice way to see
  - what is there
  - and what of that goes out on a release
but you still want the release only file to support the big current use 
case.

wspr:
@FrankMittelbach I'm still processing the best way to achieve the idea of multiple manifest files. Having a special marker is easy enough for the big list, so I'll add that presently.

For a manifest file that's send to CTAN, what would expect it to list? Just the source files listed in CTAN itself, or all the files included in the TDS? (Luckily for me, l3build creates the TDS structure regardless of whether it's included in the zip file...)

Instead of having two distinct repositories, what about having the manifest file divided into two, one for the CTAN material and one for the repo material?

wspr:
My latest commits have explored the idea of having a divided manifest, and I'm not really that sold on the idea. It looks like too much duplication, and with the naive approach I've taken (basically just listing the directory contents) you lose a lot of structure. But the upside is there are no additional options or ways to call `manifest` in the first place.

The less naive approach would be to filter out the matches based on files not in CTAN or TDS, and I think this would work except it needs more user interface.

What do you think would make the most sense here? Would you write `texlua build.lua manifest` and expect to see two manifest files, one for CTAN and one for the whole repo? Or would you think to write something like `texlua build.lua manifest` to get the "complete" version and `texlua build.lua manifest --ctan` to get the version for CTAN?

wspr:
## Time to rebase into master?

I feel like this has now been iterated enough that it would be good to rebase into master so we can address some of the other recent issues, such as splitting up the codebase into multiple lua files.

There are a few imperfections, but I think that we need more real-world use before trying to refine what's here.

Frank's suggestion that a CTAN-only manifest be part of the defaults hasn't been implemented as I'm unsure how this should look from an interface perspective. The current defaults include a repository manifest plus a CTAN manifest plus a TDS manifest all in one file, with the idea that if a user wants something more restricted they can remove whatever's unneeded.

### Squash rebase?

I'm mindful of the fact that rebasing all this into master is going to add a whole bunch of commits that aren't really that useful. Any objection to a squash rebase?

josephwright:
@wspr I'm happy with a squash: can I just have a look over the code/docs once more ...

wspr:
@josephwright — of course, take your time. I have plenty to keep me busy :-) And if you think there is still work to be done, happy to continue in this branch for the time being.

josephwright:
Function-wise looks OK.

You say in the docs that you need to do `ctan` before `manifest`. The `unpack` target is 'out of our control': `.ins` files might unpack various things. However, once that is done we can tell from the various file selections what will be installed in the TDS, typeset, _etc._ (and so what PDFs will exist). I can see that in a way it's _easier_/'safer' to pick that up from what `ctan` does as that way you don't need to duplicate the logic: is that your thinking?

I think it would be useful to have a model somewhere for how one would make a 'simple' manifest of the type I have in `siunitx` sources: just listing source and derived files, _i.e._ a manifest of what constitutes 'the work' for LPPL purposes.

The code is rather different in style to everything else: I guess you are looking after this part ;)

wspr:
@josephwright — regarding not duplicating logic, that is something I was hesitant about, and I thought down the track we could break up the main l3build logic so that we can re-use parts of it in building the manifest. Most importantly, by using the `ctan` target I knew it would be a reliable stopgap :)

Regarding *the work*, I'm more confused than ever about what people mean when they write this. My personal impression right now (which could change) is that if I'm developing a package called `fontspec` then everything in the wspr/fontspec Github repository is that work even if parts of it (such as the test suite) don't end up being distributed in CTAN.

If *the work* is only what is distributed via CTAN, then I can definitely add some more documentation about how to trim down the manifest to avoid anything else. (I'd add a user option but don't see a "clean" way to do that.) I'll treat this as the final change before merging, but I wouldn't be against various changes to logic &c. after that happens.

Finally, on style, I'm happy to adapt, especially for things like names of loop variables! Again I see that as something that we can iteratively improve...

PR #36, opened 2017-11-28
Add a sourcefiledir variable

josephwright:
This will allow structures such as that used by `beamer`, where sources and docs are separate. That is particularly useful for material *not* in `.dtx` format, where they may be significant 'structure' to the source.


wspr:
Thoughts on squashing vs not? Assuming here that rebasing is the way to go.

josephwright:
@wspr  On more 'dirs', at present I'd rather not. We can extend again if required, but I'd hope that the otehr cases are clear.

josephwright:
@wspr  I think a squash is best here: it's essentially one change on the main codebase. I'll probably do a bit of tidying up later on doc-wise: for a start, the `**` glob syntax that is also useful here isn't covered at the moment.

PR #38, opened 2017-12-07
Support for multiple test configurations (fixes #18)

### LINKED ISSUE #18: Improve approach to multiple sets of tests
Currently we have `--testfiledir|-t` but this requires each set of non-standard tests be run manually. Better would be if several test 'sets' could be run automatically. The new `call()` function should allow this with some modifications: probably something like
```lua
for _,dir in (<set-of-testfile-dirs>) do
  call(".", "check", <some-over-ride-setting-using-dir>)
end
```
where the current `--testfiledir` could be set for each call. The only issue then becomes the naming for the 'set of dirs' (or similar) function.

josephwright:
This enables multiple test configurations including saving appropriate .tlg files at the command line.

I have removed --testfiledir|-t as this seems to be redundant after the change.

Note that at present I've assumed config files are "here": we could add a configfiledir variable if desired. I will likely need to adjust this somewhat to solve the sourcefildir issue: to handle seperately!


josephwright:
Note that at present I've assumed that `.tlg` files will likely be in separate locations for separate configs, so there is no attempt to follow `<test>.<engine>.<config>.tlg`  naming.

josephwright:
I get the feeling this is probably not controversial: I'll likely merge unless there is strong objection!

PR #39, opened 2017-12-08
Enable complex source layouts (fixes #2)

### LINKED ISSUE #2: Add support for a separate doc source directory
Taken from https://github.com/dffischer/latex3/commit/f14c96ef3bc369e9d274b48c02b606ab7a1bc1b9 and the related pull request: for more complex set ups one might want a separate `docfiledir` or similar.

josephwright:
A second attempt at this: this time tested against a bundle set up too!

Various minor issues with docfiledir are fixed as part of this pull request.

Some minor issues remain concerning whether _e.g._ `textfiles` look 'here' or not. At this stage, I suggest we leave them alone and only add more vars as needed for real cases.

### DIFF
--- a/examples/Bundle-Flat/Module-One/build.lua
+++ b/examples/Bundle-Flat/Module-One/build.lua
@@ -1,0 +1,10 @@
+#!/usr/bin/env texlua
+
+bundle = "bundle-flat"
+module = "module-one"
+maindir = ".."
+
+typesetfiles  = {"*.tex"}
+
+kpse.set_program_name("kpsewhich")
+dofile(kpse.lookup("l3build.lua"))
--- a/examples/Bundle-Flat/Module-One/module-one-code.tex
+++ b/examples/Bundle-Flat/Module-One/module-one-code.tex
@@ -1,0 +1,3 @@
+
+\AtBeginDocument{\AlsoImplementation}
+\input{module-one.dtx}
--- a/examples/Bundle-Flat/Module-One/module-one.dtx
+++ b/examples/Bundle-Flat/Module-One/module-one.dtx
@@ -1,0 +1,50 @@
+%
+% \iffalse
+%<*driver>
+\ProvidesFile{module-one.dtx}
+%</driver>
+%<pkg>\ProvidesPackage{module-one}
+%<*pkg>
+  [2017/12/11 v0.1 Module One example]
+%</pkg>
+%<*driver>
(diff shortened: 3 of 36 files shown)


josephwright:
I've now assumed that `sourcefiledir` (and `docfiledir`) should be retained for CTAN uploads: this helps with something like `beamer` and is I hope a reasonable starting position.

wspr:
@josephwright I think you're right — there would be little sense (to me) to have a tree in the repository but a flat structure in CTAN. In a way, I see the CTAN upload being a snapshot of the repo itself, minus a few things that don't need to be distributed broadly (such as test files).

wspr:
In order to review this, I'd like to set up a repository of "standard" l3build example layouts. (With the idea that it might then become possible for users to copy over a layout as a template for their own package.) Where do you think it would make the most sense to do this?

* privately 
* in `wspr/l3build-examples`
* in `latex3/l3build-examples`

josephwright:
@wspr Assuming you don't feel within `l3build` itself is appropriate (`l3build/examples`), then I guess as it's likely 'official', within the `latex3` area is fine.

BTW, I've set up https://github.com/josephwright/beamer/ to use `l3build` based on this PR: you can try there seeing how the layouts match up.

We might need some minor adjustment to get all tree copying 'right', but we are I hope mostly there. I wonder if @dffischer might take a look at that.

cereda:
Code looks fine to me, I am for merging.
**Status:** GO

josephwright:
@cereda Could you open a new issue and sketch out what you mean?

cereda:
@josephwright I will take a closer look and report back. 😉

wspr:
I'm running on a Mac I don't usually use but I'm seeing something strange when running the `ctan` target. I've added an example file to demonstrate — for me, I end up with with everything seeming to run correctly, but the final `build/ctan/simple-tree/` directory is populated with two files called `code` and `doc` rather than two folders. These files appear to be the first files that should be located within the directories themselves.

Sorry I can't look into a fix for this now, it's almost midnight and I promised myself not to keep staying up too late!

josephwright:
@wspr Hopefully now fixed: `xcopy` on Windows will always copy 'correctly' here, whereas on Unix you need to `mkdir` first.

wspr:
Testing looks good 👍 I think we can merge. Would you like to do the honours?

Issue #35, opened 2017-11-10
Consider multiple files for the Lua code?


wspr:
With `l3build.lua` at around 2500 lines, I wonder if it might be time to consider a more ‘literate’ approach to the code. 

But I seem to recall that originally this was not desired since you'd then end up using l3build to unpack l3build. Do we still want to avoid having to munchhausen ourselves (as Frank would say)?

I guess an alternative might be a pretty-printing module for Lua code, so at least we could navigate the typeset code with some additional structure.

Or I might just be procrastinating from actual work, and the code is fine. Thoughts?

josephwright:
The problem with putting everything in a DTX is that you can't then use `l3build` to unpack itself: bootstrapping is then not possible. Now, one might say 'oh well' and get on with it (so you have to do `tex l3build.ins` before anything else), but this seems a bit awkward for a script that is specifically for installation and testing.

(There's the wider question that comes up from time to time of usefulness of the dtx format, particularly as most people want online docs: specifically for Lua code there is an established approach anyway ...)

wspr:
I can definitely see the argument that you don’t want to have to bootstrap yourself.

Of course we could use multiple .lua files; I don’t know if that has any downside really since we’re not loading l3build in a TeX document so we don’t need to worry about muddying up the console :)

I’d say the LaTeX world has a pretty great track record of code documentation, so I wouldn’t immediately jump to the conclusion that “established” solutions are best practice :) But I can see your point that (a) it would be nice to not have to generate the lua file(s), and (b) it would be nice to also have HTML documentation.

I’ll come back to this some time in the future I think — shouldn’t distract myself too much right now…

car222222:
> (b) it would be nice to also have HTML documentation.

Probably, but pdf is equally good for online use.

If the html “prints nicely” than there may not be much to choose between the two.

Both can be generated from a .dtx file so that way we can easily have both.

wspr:
I've just come across a scenario in which multiple files might be helpful. Basically, it would be nice to load variable defaults early so they can be used in subsequent code in a `build.lua` file. I'm imagining something like:

```
kpse.set_program_name("kpsewhich")
sourcefiles  = {"*.dtx","*.ins","unicode-math-table.tex"}
dofile("l3build-defaults.lua")
...
<more setup code which uses not only `sourcefiles` but `typesetfiles` (etc.) as well>
...
dofile("l3build.lua")
```

[...]

PR #47, opened 2018-01-06
split l3build.lua into many separate files

wspr:
see #35

no big philosophy behind the files chosen -- some may need to be divided further, others re-combined

removed some (but not all) "local" versions of function names here and there; not sure what our guiding approach should be overall

I've done some degree of testing that I've caught all the "scope" issues, although I can't claim 100% coverage


josephwright:
On the local versions business, my understanding is that Lua best practice is to always make local copies of standard stuff. For example, if you look at the ConTeXt sources they have things like
```lua
local print = print
local pairs = pairs
local sort  = table.sort
```
I can see that this makes copying code from one place to another 'interesting' (in the above, for the `sort` case but not the other two, of course).

josephwright:
I'm happy for this to get merged: it would be good to do it before dealing with any other issues ...

wspr:
Well, here goes nothing... do feel free to re-add `local` definitions of functions as you see fit! I'm not too fussed either way, but I have to admit I find it makes code harder to not have the function "prefix". OTOH if that's not best practice then so be it :)

josephwright:
@wspr We can stick with just using the global names: I'm not that fussed. (I have to admit that I find the local names a bit strange too: I have a feeling it's because I'm used to TeX.)

FrankMittelbach:
@wspr @josephwright  as I understand it this local x = x or local y = foo.y is really there for efficiency it is much faster for lua to look up a local name rather than one in its global name space and you gain a quite noticable amount of speed. However, the execution time of those scripts depend much more on the number of external (TeX) tests and the time spend there than on spending time in the actual lua code. It is therefore most likely of minor importance for us.

blefloch:
> those scripts depend much more on the number of external (TeX) tests and
> the time spend there than on spending time in the actual lua code.
Most agreed; the only exception is code that's applied to every test
file (even more so for code that's applied to every line).  Presumably
all functions there should be copied locally?

FrankMittelbach:
agreed, basically with this best practice you can gain 10% or more in speed. All I'm saying if 90% of the time spend is spend not in lua code that ends up being 1-2% in the end. On the other hand, getting into the habit of doing it always may not be a bad idea

PR #83, opened 2018-12-16
Merge in `upload` target

wspr:
I don't think this is too controversial now but I did have a couple minor sanity checks to query:

* Just to confirm, happy with `uploaddata` as the variable name of the table of metadata for CTAN? The double "d" is a little ugly but not enough to change it IMO

* I haven't documented the ability to turn off the manual verification step "do you want to upload to CTAN? [y/n]" — I think it's good to always have it there

* We currently have one variable `ctanzip` which is the ZIP file that the `ctan` target creates, AND the variable `uploadfile` which is what `upload` sends. Is this sensible or should they just be the same thing?

I've just sent through a package update to CTAN using the latest version of the code so I think it's all working as desired :)


wspr:
@FrankMittelbach — thanks for taking a look.

1. "`uploadconfig`" — better; changed

2. `ctanzip` vs `uploadfile` — agreed one variable is all that is needed; changed

josephwright:
@wspr  All looks good to me. Are you going to merge? Once you do, should I do the release or will you?

josephwright:
Oh, BTW, is the stuff in the `build.lua` file for `l3build` itself right? Doesn't seem to follow your model in the docs ...

wspr:
@josephwright — I forgot to update the material in `build.lua` itself. I guess it makes sense to keep it there, but important to make it correct!

wspr:
@josephwright — I can merge, I just thought it was good to get the all clear here first :) I'll merge tonight and if all seems well try to send to CTAN as well.

build.lua
```diff
@@ -18,7 +18,25 @@ scriptmanfiles = {"l3build.1"}
 sourcefiles  = {"*.dtx", "l3build*.lua", "*.ins"}
 typesetcmds  = "\\AtBeginDocument{\\DisableImplementation}"
 unpackdeps   = { }
-tagfiles     = {"l3build.1", "l3build.dtx", "*.md", "l3build.lua"}
+tagfiles     = {"l3build.1", "l3build.dtx", "*.md", "l3build.lua", "build.lua"}
+
+uploadconfig = {
+  version     = "2018-12-17",
```
josephwright: This works OK at present for development/experimental status, but I'd hope we can move to not having the data in the build script ... probably should have a function to extract it automatically from sources and/or Git/Subversion/... Something like `function get_tag_data()`?

wspr: Indeed. One part of me thinks that if it's going to be hardcoded anywhere then the build.lua file should be it! I also like the idea of only "filling in" the data when `install` or `ctan` is run (but that only works for dates).

Like you say, this works for now...

wspr:
Well, that's done and sent off to CTAN! If there aren't any problems (touch wood) we can delete the branch.

PR #107, opened 2019-10-31
Normalize lua data references

zauguin:
Some values contained in nodes are represented by Lua values in LuaTeX. This leds LuaTeX to print log output like
```
....\pdfliteral page <lua data reference 792>
```
where the number is highly version dependent and not meaningful. I suggest normalizing these to `reference ...>`.

### DIFF
--- a/l3build-check.lua
+++ b/l3build-check.lua
@@ -251,6 +251,9 @@ local function normalize_log(content,engine,errlevels)
     line = gsub(line, "save cache:", "load cache:")
     -- A tidy-up to keep LuaTeX and other engines in sync
     line = gsub(line, utf8_char(127), "^^?")
+    -- Remove lua data reference ids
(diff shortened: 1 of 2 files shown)


josephwright:
Looks good: could you add something to the docs about this too?

zauguin:
@josephwright I added a short note.

car222222:
@zauguin  I think your auto-correct is responsible for this note starting with: “Conversation . . .” !!
But in fact you are just dropping the number so maybe the note should be more like: “These are normalized by removing the (lua( internal) number. “

zauguin:
Fixed the typo.

> But in fact you are just dropping the number so maybe the note should be more like: “These are normalized by removing the (lua( internal) number. “

I thought about that but I wanted to keep it consistent: The previous four points all talk about replacing some number by `...` and all use "Conversion", so suddenly using another formulation would give the misleading impression that there is something different going on.

PR #109, opened 2020-02-09
Avoid temporary file when passing "y" in unpack

zauguin:
Mostly I just don't like temporary files and I interpreted the statement "as it's the only way on Windows" as a challenge.
Also this avoids the need for a system-dependent `os_yes` command.

This PR does not delete the `os_yes` variable to keep compatibility with build.lua files which might rely on it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ this project uses date-based 'snapshot' version identifiers.
 ### Changed
 - Avoid temporary file when unpacking
 
+### Deprecated
+- `os_yes`: use `io.popen(...,w)` instead
+
 ## [2020-02-17]
 
 ### Added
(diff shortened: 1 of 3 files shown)


josephwright:
Code looks OK when run against the LaTeX2e repo, so I wonder why the tests fail! 

I'd remove `os_yes`, etc.: it's a purely internal thing.

zauguin:
> Code looks OK when run against the LaTeX2e repo, so I wonder why the tests fail!

I wonder too. Locally the tests work fine and the branch and PR builds (https://travis-ci.org/latex3/l3build/builds/647887366?utm_medium=notification&utm_source=github_status and https://travis-ci.org/latex3/l3build/jobs/647886221?utm_medium=notification&utm_source=github_status) show completely different errors.

> 
> I'd remove `os_yes`, etc.: it's a purely internal thing.

Currently `os_yes` is also used in LaTeX2e `base/build.lua`, so I think it should stay at least until a similar change has been applied there.

PR #128, opened 2020-10-22
Allow additional test types

zauguin:
Add support for custom test types with individual normalization/processing and potentially independent comparision code.

### DIFF
--- a/l3build-check.lua
+++ b/l3build-check.lua
@@ -223,19 +223,19 @@ local function normalize_log(content,engine,errlevels)
       -- Remove '\displace 0.0' lines in (u)pTeX
       if match(line,"^%.*\\displace 0%.0$") then
         return ""
-       end
-     end
+      end
+    end
     -- Deal with Lua function calls
     if match(line, "^Lua function") then
       line = gsub(line,"= %d+$","= ...")
     end
-     -- Remove the \special line that in DVI mode keeps PDFs comparable
+    -- Remove the \special line that in DVI mode keeps PDFs comparable
     if match(line, "^%.*\\special%{pdf: docinfo << /Creator") or
       match(line, "^%.*\\special%{ps: /setdistillerparams") or
       match(line, "^%.*\\special%{! <</........UUID") then
       return ""
     end
-     -- Remove \special lines for DVI .pro files
+    -- Remove \special lines for DVI .pro files
     if match(line, "^%.*\\special%{header=") then
       return ""
     end
@@ -255,7 +255,7 @@ local function normalize_log(content,engine,errlevels)
        match(line, "^used file       >") or
        match(line, "^used option     >") or
        match(line, "^used structure  >") then
-       return ""
+      return ""
     end
     -- The first time a new font is used by LuaTeX, it shows up
     -- as being cached: make it appear loaded every time
@@ -374,20 +374,20 @@ local function normalize_lua_log(content,luatex)
     -- This block only applies to the output of LuaTeX itself,
     -- hence needing a flag to skip the case of the reference log
     if luatex and
-       tonumber(luatex_version) >= 107 and
-       match(line, "^%.*\\kern") then
-       -- Re-insert the space in explicit kerns
-       if match(line, "kern%-?%d+%.%d+ *$") then
-         line = gsub(line, "kern", "kern ")
-       elseif match(line, "%(accent%)$") then
-         line = gsub(line, "kern", "kern ")
-         line = gsub(line, "%(accent%)$", "(for accent)")
-       elseif match(line, "%(italic%)$") then
-         line = gsub(line, "kern", "kern ")
-         line = gsub(line, " %(italic%)$", "")
-       else
-         line = gsub(line, " %(font%)$", "")
-       end
+      tonumber(luatex_version) >= 107 and
+      match(line, "^%.*\\kern") then
+      -- Re-insert the space in explicit kerns
+      if match(line, "kern%-?%d+%.%d+ *$") then
+        line = gsub(line, "kern", "kern ")
+      elseif match(line, "%(accent%)$") then
+        line = gsub(line, "kern", "kern ")
+        line = gsub(line, "%(accent%)$", "(for accent)")
+      elseif match(line, "%(italic%)$") then
+        line = gsub(line, "kern", "kern ")
+        line = gsub(line, " %(italic%)$", "")
+      else
+        line = gsub(line, " %(font%)$", "")
+      end
     end
     -- Changes in PDF specials
     line = gsub(line, "\\pdfliteral origin", "\\pdfliteral")
@@ -398,8 +398,8 @@ local function normalize_lua_log(content,luatex)
     -- 'Recover' some discretionary data
     if match(lastline, "^%.+\\discretionary %(penalty 50%)$") and
        match(line, boxprefix(lastline) .. "%.= ") then
-       line = gsub(line," %(font%)$","")
-       return gsub(line, "%.= ", ""),""
+      line = gsub(line," %(font%)$","")
+      return gsub(line, "%.= ", ""),""
     end
     -- Where the last line was a discretionary, looks for the
     -- info one level in about what it represents
@@ -410,13 +410,13 @@ local function normalize_lua_log(content,luatex)
       local prefix = boxprefix(lastline)
       if match(line, prefix .. "%.") or
          match(line, prefix .. "%|") then
-         if match(lastline, " replacing $") and
-            not dropping then
-           -- Modify the return line
-           return gsub(line, "^%.", ""), lastline, true
-         else
-           return "", lastline, true
-         end
+        if match(lastline, " replacing $") and
+           not dropping then
+          -- Modify the return line
+          return gsub(line, "^%.", ""), lastline, true
+        else
+          return "", lastline, true
+        end
       else
         if dropping then
           -- End of a \discretionary block
@@ -465,7 +465,7 @@ local function normalize_lua_log(content,luatex)
        match(line, "^%.+\\localbrokenpenalty=0$")    or
        match(line, "^%.+\\localleftbox=null$")       or
        match(line, "^%.+\\localrightbox=null$")      then
-       return "", ""
+      return "", ""
     end
     -- Older LuaTeX versions set the above up as a whatsit
     -- (at some stage this can therefore go)
@@ -526,7 +526,7 @@ local function normalize_pdf(content)
         if binary then
           new_content = new_content .. "[BINARY STREAM]" .. os_newline
         else
-           new_content = new_content .. stream_content .. line .. os_newline
+          new_content = new_content .. stream_content .. line .. os_newline
         end
         binary = false
       else
@@ -554,10 +554,19 @@ local function normalize_pdf(content)
   return new_content
 end
 
+function rewrite_log(source, result, engine, errlevels)
+  return rewrite(source, result, normalize_log, engine, errlevels)
+end
+
+function rewrite_pdf(source, result, engine, errlevels)
+  return rewrite(source, result, normalize_pdf, engine, errlevels)
+end
+
 -- Run one test which may have multiple engine-dependent comparisons
 -- Should create a difference file for each failed test
 function runcheck(name, hide)
-  if not testexists(name) then
+  local test_filename = testexists(name)
+  if not test_filename then
     print("Failed to find input for test " .. name)
     return 1
   end
@@ -566,9 +575,9 @@ function runcheck(name, hide)
     checkengines = options["engine"]
   end
   -- Used for both .lvt and .pvt tests
-  local function check_and_diff(ext,engine,comp,pdftest)
-    runtest(name,engine,hide,ext,pdftest,true)
-    local errorlevel = comp(name,engine)
+  local function check_and_diff(test_type, engine)
+    runtest(name, engine, hide, test_type.test, test_type, true)
+    local errorlevel = test_type.compare(name,engine)
     if errorlevel == 0 then
       return errorlevel
     end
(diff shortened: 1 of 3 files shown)


josephwright:
Is this a 'work in progress' or do you want comments?

zauguin:
I would appreciate comments, especially about the choice of customization points.  (Sorry for the whitespace changes, I normalized the file in the end  and that picked up some old lines too, but it felt silly to revert that.)

l3build-check.lua
```diff
@@ -554,10 +554,19 @@ local function normalize_pdf(content)
...
 end
 
+function rewrite_log(source, result, engine, errlevels)
+  return rewrite(source, result, normalize_log, engine, errlevels)
+end
+
+function rewrite_pdf(source, result, engine, errlevels)
+  return rewrite(source, result, normalize_pdf, engine, errlevels)
+end
+
```
zauguin, line 564: This might look a bit odd because there already is `rewrite` which can be customized with a normalize function, for for other formats it is often easier to access a file directly, therefore this additional indirection.

l3build-check.lua
```diff
@@ -602,64 +613,73 @@ end
...
+    return compare(difffile, reffile, genfile, cleanup, name, engine)
+  end
   local errorlevel = execute(os_diffexe .. " "
-    .. normalize_path(tpffile .. " " .. pdffile .. " > " .. difffile))
+    .. normalize_path(reffile .. " " .. genfile .. " > " .. difffile))
   if errorlevel == 0 or cleanup then
     remove(difffile)
   end
   return errorlevel
 end
```
zauguin, line 678: One of the bigger customization points. The idea is that this will almost always just be diff which is therefore the default, but it allows more flexibility if necessary. The base_compare wrapper ensures that the interface is relatively simple. In most cases, a compare function can just take three arguments with the output and the to be compared input filenames.

l3build-check.lua
```diff
@@ -691,7 +711,7 @@ end
 
 -- Run one of the test files: doesn't check the result so suitable for
 -- both creating and verifying
-function runtest(name, engine, hide, ext, pdfmode, breakout)
+function runtest(name, engine, hide, ext, test_type, breakout)
```
zauguin, line 714: It would be a bit nicer to hide `ext` too in `test_type`, but it has to be separate to allow expectation files.

l3build-check.lua
```diff
@@ -777,47 +795,34 @@ function runtest(name, engine, hide, ext, pdfmode, breakout)
...
-           if fileexists(testdir,newname) then
-             rm(testdir,newname)
-           end
-           ren(testdir,file,newname)
+      if match(file,"^" .. name .. "%.[^.]+$") then
+        local newname = gsub(file,"(%.[^.]+)$","." .. engine .. "%1")
+        if fileexists(testdir,newname) then
+          rmfile(testdir,newname)
         end
+        ren(testdir,file,newname)
```
zauguin, line 825: This whole loop could be made faster by skipping the directory listing and just checking the filenames directly, but then we woul no longer allow arbitrary patterns in auxfiles so this would be a breaking change.

l3build-check.lua
```diff
@@ -777,47 +795,34 @@ function runtest(name, engine, hide, ext, pdfmode, breakout)
...
+  test_type.rewrite(gen_file,new_file,engine,errlevels)
   -- Store secondary files for this engine
   for _,filetype in pairs(auxfiles) do
     for _,file in pairs(filelist(testdir, filetype)) do
-      if match(file,"^" .. name .. ".[^.]+$") then
-        local ext = match(file, "%.[^.]+$")
-        if ext ~= lvtext and
-           ext ~= tlgext and
-           ext ~= lveext and
-           ext ~= logext then
```
zauguin, line 814: This drops the explicit exclusion of out types, but they really shouldn't be matched as auxfiles anyway.

l3build-check.lua
```diff
@@ -848,36 +857,42 @@ function check(names)
...
+              end
+            end
+            if not exclude then
+              if fileexists(testfiledir .. "/" .. name) then
+                return 1
+              end
+              insert(names,jobname(name))
+            end
           end
-          addname(name)
```
zauguin, line 880: Here we change the logic in the process: Instead of first listing all excluded files, we check for exclusions while scanning. That's a bit cleaner to implement in the generic case and it also saves some directory listings.

l3build-check.lua
```diff
@@ -992,32 +1007,36 @@ function save(names)
...
+          end
         end
-        if errorlevel ~=0 then return errorlevel end
+        if errorlevel ~= 0 then return errorlevel end
       end
     elseif locate({unpackdir, testfiledir}, {name .. lveext}) then
+      -- FIXME: This doesn't look like it does what it claims to do.
       print("Saved " .. tlgext .. " file overrides a "
         .. lveext .. " file of the same name")
       return 1
```
zauguin, line 1042: I left this error as is for now because I didn't understand what it's supposed to do. It errors if no test file, but an expectation file exists. But why is it talking about overwriting? If I'm not missing something, the saved tlgfile can't overwrite anything because without the testfile we can't even generate a tlg file.

l3build-file-functions.lua
```diff
@@ -49,75 +49,22 @@ local gsub             = string.gsub
...
+           + '*' * l.Cc'.*'
+           + '\\' * (-1 + l.C(escaped_char)))
+      + l.P'[' / function()
+            print("[...] syntax not supported in globs!")
+            return ''
+         end
+      + escaped_char)^0 * l.Cc'$')
+   function glob_to_pattern(glob)
+      return to_pattern:match(glob)
+   end
```
zauguin, line 67: I just wanted shorter patterns during debugging so I rewrote the converter and committed it by accident. I will revert this before merging.

l3build-variables.lua
```diff
@@ -209,6 +209,24 @@ pvtext = pvtext or ".pvt"
...
+   },
+   pdf = {
+      test = pvtext,
+      generated = pdfext,
+      reference = tpfext,
+      rewrite = rewrite_pdf,
+   },
+}})
+test_order = test_order or {"log", "pdf"}
+
```
zauguin, line 229: The default configuration. Maybe three interesting parts here: The split between `test_types` and `test_order`. Mostly it's done this way to ensure that behavior is deterministic in some edge cases, but it also allows to a switch to a auto loading schemes for additional formats later (To automatically fill test_types when types are used in test_order).

Names might need improvement.

zauguin (commented):
Some inline comments bout the more interesting parts. (Here more interesting means: Not just replacing some variable names or replacing repetition with loops)

l3build-check.lua
```diff
@@ -583,13 +592,15 @@ function runcheck(name, hide)
...
+    local errlevel
+    for _, kind in ipairs(test_order) do
+      local test_type = test_types[kind]
+      if test_filename == testfiledir .. "/" .. name .. test_type.test then
+        errlevel = check_and_diff(test_type, engine)
+        break
+      end
     end
-    if errlevel ~= 0 and options["halt-on-error"] then
+    if assert(errlevel) ~= 0 and options["halt-on-error"] then
```
josephwright, line 603: Do we really want an assertion here? Won't that leave us with an uncontrolled error?

l3build-check.lua
josephwright: Probably this was historical anyway: it looks OK but we need to see what happens when applied to the 2e repo :)

l3build-check.lua
josephwright: This was written a while ago when things were rather less complicated: I may well have copied the text between two places where the outcomes where different! I agree it doesn't look like it doesn't seem to make sense: change the text to say that?

l3build-file-functions.lua
josephwright: Feel free to change if it works better: I simply took some existing code, simplified a little and used it. I'm not really good with LPEG so it's very likely that there are better ways.

l3build-variables.lua
josephwright: I'm not clear on what you are up to with metatables here :) Why not `test_types = test_types or  { }` then 
`test_types.log = test_types.log or { ... }` and so on?

l3build.dtx
```diff
@@ -1246,6 +1248,38 @@
...
+%   mytest = {
+%     test = ".mylvt",
+%     reference = ".mytlg",
+%     generated = ".log",
+%     rewrite = function(source, normalized, engine, errorcode)
+%       -- In this example we just copy the logfile without any normalization
+%       os.execute(string.format("cp %s %s", source, normalized)
+%     end,
+%   },
+% }
```
josephwright, line 1280: This does not match how everything else works. I'd expect `test_types  =  test_types or { }`, `test_types.mytest = test_types.mytext or { ... }`, the same as we do for 'special' engines and so on.

josephwright: What happens if only some entries are given? Do we need a prototype to be copied then altered, for example?

josephwright (changes requested):
Looks good overall but there are a few places I'd like a bit of clarification (no code changes required necessarily).

l3build-check.lua
zauguin: 2e does not change auxfiles, therefore ext can never have these values anyway.

l3build-check.lua
zauguin: I would just drop this branch. It only triggers when a `lve` file exists but no test and then `l3build save` is called. While the `lve` file is odd, the real issue we should be reporting is the missing test.

l3build.dtx
zauguin: I don't really understand what you mean here.

All the entries `test`, `reference` and `generated` are required because otherwise the system doesn't know which files to operate on. Similarly, `rewrite` is effectivly always required because there is no reasonable default (We could copy without normalization but that would never be reproducible.)

The case of copying and altering a prototype sounds like you would just have two parts in `rewrite`: First copy, then rewrite. (Or both steps can be combined as in the default normalization functions)

l3build-check.lua
zauguin: It doesn't do anything because `errlevel` can never be `nil` here. (I had the assert to ensure that I didn't miss a case where it can happen, but I'll take it out again.

l3build-check.lua
zauguin: @josephwright The new version now moves the error to actually warn if a lve is present while trying to save a test.

l3build-variables.lua
zauguin: @josephwright No real reason, I'm just more used to the metatable approach. I changed it for consistency now.

josephwright:
I think modulo a few merge issues we are good to go here. I think also that with one or two more tweaks this would allow @jlaurens to test Lua - we just need to make sure that `\input` is not added in those cases, which we can do by adding some entry to the setup table.

Anyone else got comments (@cereda ?)

jlaurens:
The reference/comparison model does not fit with unit testing without a very very significant overhead. Some work must be done before attempting to use this route, so do not feel in a hurry for me. This types enhancement makes sense on its own.

In my opinion, the main remark is that unit testing has nothing to copy to the build directory.
So there is no reference, no expectation, no normalisation and nothing to rewrite. The only thing that makes a difference seems to be the `compare` method. I have difficulties to imagine a straightforward workflow for this approach.

However, the build directory might be used as a playground to test file functions on temp files and folders.

BTW, some how to recipe may help in following the path, but that is not a requirement of mine.

josephwright:
> The reference/comparison model does not fit with unit testing without a very very significant overhead. Some work must be done before attempting to use this route, so do not feel in a hurry for me. This types enhancement makes sense on its own.

Indeed: I'd mainly got distracted from looking over the PR here. It's a great idea, we just have be sure we've got the interface about right. I plan to check that partly by setting up some Lua testing in a dependent branch, and making sure it looks OK.

> In my opinion, the main remark is that unit testing has nothing to copy to the build directory.

I'm afraid I don't see that. The reason we move everything for testing is that is the only way to have an 'isolated' system in which one can be sure there are not stray accessible files. Copying everything to a working space ensures there are no 'oddities'. At the same time, other than `l3build` itself and a few special cases (`luaotfload` is the obvious one), I'd expect Lua projects in the 'TeX sphere' to be packaged as `.dtx` sources or similar, so there's the need to extract using DocStrip first. That's a lot cleaner done in a 'disposable' space.

> So there is no reference, no expectation, no normalisation and nothing to rewrite. The only thing that makes a difference seems to be the `compare` method. I have difficulties to imagine a straightforward workflow for unit testing with this approach.

Seems simple to me. Write a test file (could be as `.lua`), output the results to a file (say `.log`), save a standardised version as say `.llg`. So more-or-less the same as for LaTeX, just with different extensions. That might be supported by some `regression-test.lua` in the style of `regression-test.tex`, but could be a lot less complex (no real need to worry about normalisation after-the-fact in most cases).

jlaurens:
You are still far ahead of me. You are more talking about adding new regression testing capabilities to l3build whereas I am focused on developing l3build with the help of unit testing.
So your examples do make sense from your purpose but not for mine.

josephwright:
> You are still far ahead of me. You are more talking about adding new regression testing capabilities to l3build whereas I am focused on developing l3build with the help of unit testing.
> So your examples do make sense from your purpose but not for mine.

Huh? You can happily do something akin to
```lua
require("some-module")

if some.thing then
  TESTOUT("All good")
else
   TESTOUT("Fail")
end
```
using some `TESTOUT` that writes to e.g. a `.log` file.

jlaurens:
simple on the paper. On everyday use, one gets rapidly into trouble.
Not to mention the testing aspect that must be practical (no intermediate text file, simple code), just trying to
```
require `l3build`
```
is actually difficult. One of these is required upwind
- redefine main
- define global options and bypass option parsing
- pass options to the testing script
- use os.execute with options instead of require
That sounds a bit complicated for a testing technique. Anyway, it does not give fine grain control.

I easily imagine that testing "arguments", "variables" or "file functions" module alone is rather easy.
To test "help", both "arguments" and "stdmain" are required but I guess the tests would be ok without the rest.
For the others, I am less optimistic because the tests must be made in a global context.
This global context comes from arguments parsing and variables management.
Both can be simulated in theory, but does It worth the cost?

josephwright:
@zauguin  Could you put in the LPEG stuff as a separate PR? That would make things nice and clean in many ways, but is separate from the main part here.

zauguin:
@josephwright Done. I also used the this to rebase on current master.

josephwright:
@zauguin  Er, don't see a separate PR!

zauguin:
@josephwright Any other comments or can we merge this after resolving the conflicts?

josephwright:
I think we are good to go

PR #129, opened 2021-01-08
Clarify the documentation of targets tag and upload

jlaurens:
Optional arguments are added, some renaming and rephrasing to be more explicit and more consistent.

### DIFF
--- a/l3build.dtx
+++ b/l3build.dtx
@@ -340,11 +340,10 @@
 % \item install
 % \item manifest
 % \item save \meta{name(s)}
-% \item tag \meta{tag}
+% \item tag [\meta{tag name}]
 % \item uninstall
 % \item unpack
-% \item upload
-% \item upload \meta{tag}
+% \item upload [\meta{version}]
 % \end{itemize}
 % These targets are described below.
 %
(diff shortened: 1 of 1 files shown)


l3build.dtx
```diff
@@ -340,11 +340,10 @@
 % \item install
 % \item manifest
 % \item save \meta{name(s)}
-% \item tag \meta{tag}
+% \item tag [\meta{tag name}]
 % \item uninstall
 % \item unpack
-% \item upload
-% \item upload \meta{tag}
+% \item upload [\meta{version}]
```
josephwright, line 346: I don't feel the tag is optional!

FrankMittelbach: is it not @josephwright ? In my own repositories I used upload in the past without any tag (wasn't even aware that one can give it one, but good to know.

FrankMittelbach: but I guess in any case it should be called "tag" not "version"  or am I totally misunderstand what it does?

jlaurens: According to the `upload()` [https://github.com/latex3/l3build/blob/ef103b1fb6960e0981d034ab935155dfcc12baec/l3build-upload.lua#L103](source code),
the argument is optional and defaults to `uploadconfig.version` as mentioned below (hence the "version" instead of "tag")

Moreover `l3build upload` does not report any error.

josephwright: OK, I was obviously wrong: I thought it was needed for CTAN.

FrankMittelbach: ah, then perhaps that need even more documentation as it is about the CTAN version of the file. That's actually quite helpful if that can be set on the command line

PR #130, opened 2021-01-13
Reference to https://www.ctan.org/upload

jlaurens:
The actual documentation refers to the CTAN upload API which the end user won't use directly.
Instead explicitly referring to the web page is extremely useful and efficient.

BTW, ignoring auxiliary files allows to typeset the dtx file from a standard tex editor without messing up the repository.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,11 +3,3 @@ build/
 *.zip
 !testfiles/*.pdf
 
-l3build.glo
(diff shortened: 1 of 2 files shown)


josephwright:
Could you update so we just have the 'business end' here: I think the `.gitignore` part should be separate (and we can discuss the utility ....).

PR #132, opened 2021-01-13
turning l3build-arguments.lua into a module

jlaurens:
The lua code in the repository really needs some serious cleaning. Actually, it is rather hard to understand because quite everything is at the top level. Here are related problems

some of the globals should be initialized before use
some globals should turn to local because they are only used in a very limited scope
ctan_post should not be at the same time global and local.
some locals are useless shallow copies of globals
The first goal is to organize the code into more separated modules with a well defined dependency.
The second goal will be to have some regression tests.

In that pull request, l3build-arguments.lua is turned into a module. All inside references to global variables are turned into formal parameters references. Now the l3build.lua main controller creates the global variable named Opts to replace the former variable that was created inside l3build-arguments.lua as a side effect.

All options["foo"] have been renamed to Opts.foo for a much better legibility.

### DIFF
--- a/l3build-arguments.lua
+++ b/l3build-arguments.lua
@@ -21,8 +21,8 @@ The development version of the bundle can be found at
 for those people who are interested.
 
 --]]
+local A = {}
 
-local exit             = os.exit
 local stderr           = io.stderr
 
 local find             = string.find
@@ -34,7 +34,7 @@ local insert           = table.insert
 
 -- Parse command line options
 
-option_list =
+A.option_list =
   {
     config =
       {
@@ -158,39 +158,38 @@ option_list =
 
 -- This is done as a function (rather than do ... end) as it allows early
 -- termination (break)
-local function argparse()
-  local result = { }
-  local names  = { }
-  local long_options =  { }
-  local short_options = { }
+A.argparse = function (arg)
+  local result = {}
+  local names  = {}
+  local long_options =  {}
+  local short_options = {}
   -- Turn long/short options into two lookup tables
-  for k,v in pairs(option_list) do
-    if v["short"] then
-      short_options[v["short"]] = k
+  for k,v in pairs(A.option_list) do
+    if v.short then
+      short_options[v.short] = k
     end
     long_options[k] = k
   end
-  local args = args
   -- arg[1] is a special case: must be a command or "-h"/"--help"
   -- Deal with this by assuming help and storing only apparently-valid
   -- input
   local a = arg[1]
-  result["target"] = "help"
+  result.target = "help"
   if a then
     -- No options are allowed in position 1, so filter those out
     if a == "--version" then
-      result["target"] = "version"
+      result.target = "version"
     elseif not match(a, "^%-") then
-      result["target"] = a
+      result.target = a
     end
   end
   -- Stop here if help or version is required
-  if result["target"] == "help" or result["target"] == "version" then
+  if result.target == "help" or result.target == "version" then
     return result
   end
   -- An auxiliary to grab all file names into a table
   local function remainder(num)
-    local names = { }
+    local names = {}
     for i = num, #arg do
       insert(names, arg[i])
     end
@@ -235,7 +234,7 @@ local function argparse()
       local optname = opts[opt]
       if optname then
         -- Tidy up arguments
-        if option_list[optname]["type"] == "boolean" then
+        if A.option_list[optname].type == "boolean" then
           if optarg then
             local opt = "-" .. (match(a, "^%-%-") and "-" or "") .. opt
             stderr:write("Value not allowed for option " .. opt .."\n")
@@ -257,10 +256,10 @@ local function argparse()
       end
       -- Store the result
       if optarg then
-        if option_list[optname]["type"] == "string" then
+        if A.option_list[optname].type == "string" then
           result[optname] = optarg
         else
-          local opts = result[optname] or { }
+          local opts = result[optname] or {}
           for hit in gmatch(optarg, "([^,%s]+)") do
             insert(opts, hit)
           end
@@ -277,22 +276,20 @@ local function argparse()
     end
   end
   if next(names) then
-   result["names"] = names
+   result.names = names
   end
   return result
 end
 
-options = argparse()
-
 -- Sanity check
-function check_engines()
-  if options["engine"] and not options["force"] then
-     -- Make a lookup table
-     local t = { }
+A.check_engines = function (options, checkengines)
+  if options.engine and not options.force then
+    -- Make a lookup table
+    local t = {}
     for _, engine in pairs(checkengines) do
       t[engine] = true
     end
-    for _, engine in pairs(options["engine"]) do
+    for _, engine in pairs(options.engine) do
       if not t[engine] then
         print("\n! Error: Engine \"" .. engine .. "\" not set up for testing!")
         print("\n  Valid values are:")
@@ -300,8 +297,10 @@ function check_engines()
           print("  - " .. engine)
         end
         print("")
-        exit(1)
+        os.exit(1)
       end
     end
   end
 end
+
+return A
--- a/l3build-aux.lua
+++ b/l3build-aux.lua
@@ -68,7 +68,7 @@ end
 -- Do some subtarget for all modules in a bundle
 function call(dirs, target, opts)
   -- Turn the option table into a string
-  local opts = opts or options
+  local opts = opts or Opts
   local s = ""
   for k,v in pairs(opts) do
     if k ~= "names" and k ~= "target" then -- Special cases
--- a/l3build-aux.lua
+++ b/l3build-aux.lua
@@ -72,7 +72,7 @@ function call(dirs, target, opts)
   local s = ""
   for k,v in pairs(opts) do
     if k ~= "names" and k ~= "target" then -- Special cases
-      local t = option_list[k] or { }
+      local t = A.option_list[k] or { }
       local arg = ""
       if t["type"] == "string" then
         arg = arg .. "=" .. v
--- a/l3build-check.lua
+++ b/l3build-check.lua
@@ -55,7 +55,7 @@ local remove           = os.remove
 -- Set up the check system files: needed for checking one or more tests and
 -- for saving the test files
 function checkinit()
-  if not options["dirty"] then
+  if not Opts.dirty then
     cleandir(testdir)
     cleandir(resultdir)
   end
@@ -562,8 +562,8 @@ function runcheck(name, hide)
     return 1
   end
   local checkengines = checkengines
-  if options["engine"] then
-    checkengines = options["engine"]
+  if Opts.engine then
+    checkengines = Opts.engine
   end
   -- Used for both .lvt and .pvt tests
   local function check_and_diff(ext,engine,comp,pdftest)
@@ -572,10 +572,10 @@ function runcheck(name, hide)
     if errorlevel == 0 then
       return errorlevel
     end
-    if options["show-log-on-error"] then
+    if Opts["show-log-on-error"] then
       showfailedlog(name)
     end
-    if options["halt-on-error"] then
+    if Opts["halt-on-error"] then
       showfaileddiff()
     end
     return errorlevel
@@ -589,7 +589,7 @@ function runcheck(name, hide)
     else
       errlevel = check_and_diff(lvtext,engine,compare_tlg)
     end
-    if errlevel ~= 0 and options["halt-on-error"] then
+    if errlevel ~= 0 and Opts["halt-on-error"] then
       return 1
     end
     if errlevel > errorlevel then
@@ -838,7 +838,7 @@ end
 function check(names)
   local errorlevel = 0
   if testfiledir ~= "" and direxists(testfiledir) then
-    if not options["rerun"] then
+    if not Opts.rerun then
       checkinit()
     end
     local hide = true
@@ -882,10 +882,10 @@ function check(names)
       end
       sort(names)
       -- Deal limiting range of names
-      if options["first"] then
+      if Opts.first then
         local allnames = names
         local active = false
-        local firstname = options["first"]
+        local firstname = Opts.first
         names = { }
         for _,name in ipairs(allnames) do
           if name == firstname then
@@ -896,9 +896,9 @@ function check(names)
           end
         end
       end
-      if options["last"] then
+      if Opts.last then
         local allnames = names
-        local lastname = options["last"]
+        local lastname = Opts.last
         names = { }
         for _,name in ipairs(allnames) do
           insert(names,name)
@@ -917,7 +917,7 @@ function check(names)
       end
       return tbl
     end
-    if options["shuffle"] then
+    if Opts.shuffle then
       names = shuffle(names)
     end
     -- Actually run the tests
@@ -929,7 +929,7 @@ function check(names)
       local errlevel = runcheck(name, hide)
       -- Return value must be 1 not errlevel
       if errlevel ~= 0 then
-        if options["halt-on-error"] then
+        if Opts["halt-on-error"] then
           return 1
         else
           errorlevel = 1
@@ -986,7 +986,7 @@ end
 
 function save(names)
   checkinit()
-  local engines = options["engine"] or {stdengine}
+  local engines = Opts.engine or {stdengine}
   if names == nil then
     print("Arguments are required for the save command")
     return 1
--- a/l3build-install.lua
+++ b/l3build-install.lua
@@ -37,13 +37,13 @@ local insert = table.insert
 
 local function gethome()
   set_program("latex")
-  return abspath(options["texmfhome"] or var_value("TEXMFHOME"))
+  return abspath(Opts.texmfhome or var_value("TEXMFHOME"))
 end
 
 function uninstall()
   local function zapdir(dir)
     local installdir = gethome() .. "/" .. dir
-    if options["dry-run"] then
+    if Opts["dry-run"] then
       local files = filelist(installdir)
       if next(files) then
         print("\n" .. "For removal from " .. installdir .. ":")
@@ -72,7 +72,7 @@ function uninstall()
       -- Man files should have a single-digit extension: the type
       local installdir = gethome() .. "/doc/man/man"  .. match(file,".$")
       if fileexists(installdir .. "/" .. file) then
-        if options["dry-run"] then
+        if Opts["dry-run"] then
           insert(manfiles,"man" .. match(file,".$") .. "/" ..
            select(2,splitpath(file)))
         else
@@ -294,5 +294,5 @@ function install_files(target,full,dry_run)
 end
 
 function install()
-  return install_files(gethome(),options["full"],options["dry-run"])
+  return install_files(gethome(),Opts.full,Opts["dry-run"])
 end
--- a/l3build.lua
+++ b/l3build.lua
@@ -32,7 +32,6 @@ local lfs = require("lfs")
 
 -- Local access to functions
 
-local assert           = assert
 local ipairs           = ipairs
 local insert           = table.insert
 local lookup           = kpse.lookup
@@ -40,19 +39,17 @@ local match            = string.match
 local gsub             = string.gsub
 local next             = next
 local print            = print
-local select           = select
-local tonumber         = tonumber
 local exit             = os.exit
 
 -- l3build setup and functions
 kpse.set_program_name("kpsewhich")
-build_kpse_path = match(lookup("l3build.lua"),"(.*[/])")
+local build_kpse_path = match(lookup("l3build.lua"),"(.*[/])")
 local function build_require(s)
-  require(lookup("l3build-"..s..".lua", { path = build_kpse_path } ) )
+  return require(lookup("l3build-"..s..".lua", { path = build_kpse_path } ) )
 end
 
 -- Minimal code to do basic checks
-build_require("arguments")
+local A = build_require("arguments")
 build_require("help")
 
 build_require("file-functions")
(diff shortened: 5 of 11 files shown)


blefloch:
Side remark not related to code: you should seriously have omitted your first sentence and a half, whose sole effect is to insult whomever wrote this code.

Regarding one of the technical points "some locals are useless shallow copies of globals": as far as I remember this was for efficiency: accessing these shallow copies might be faster in some contexts, and it was extended to most functions for stylistic consistency.

Some of your other points seem easy to improve and worth it: ctan_post both global and local; initialization of all globals; turning some globals into locals.

I think splitting the code into modules was considered (there may be a branch by @wspr somewhere), but wasn't worth the effort, since it is not clear who would use the building blocks on their own.  Adding regression tests would be nice; there are a few in the `testfiles*` directories.

josephwright:
A few comments before getting to any code detail ;)

> The lua code in the repository really needs some serious cleaning. Actually, it is rather hard to understand because quite everything is at the top level.

I think it's worth noting that `l3build` grew out of basically about a dozen lines of code for replacing an older Perl script, and it's the first (and thus-far only) large-ish piece of Lua code I've written. I'm also a TeX programmer, so I tend toward a certain approach: for example, I am more comfortable with global variables than perhaps a typical Lua programmer is.

When @wspr split the single `l3build.lua` source into several files, we did a bit of cleanup. However, I have no doubt there is a lot more we could do. It's partly a question of priority: a rewrite would be in some ways ideal, but as the main job of `l3build` is to do LaTeX releases, and that works, we have other things in our minds.

I'll probably as @zauguin and @paulocereda for comment: they are my go-to Lua experts.

> some of the globals should be initialized before use

I guess this might make more sense when I take a look at the code.

> some globals should turn to local because they are only used in a very limited scope

Quite possibly, as long as they are not used outside @latexproject code.

> ctan_post should not be at the same time global and local.

Quite possibly: that one I'm not sure about.

> some locals are useless shallow copies of globals

I wonder if @blefloch is right here: are you talking _e.g._ `local print = print`?

> The second goal will be to have some regression tests.

I'm not disagreeing, but just to note that the definitive question always is 'does a change to `l3build` show up when running `l3build ctan` in the LaTeX2e repo'.

> In that pull request, l3build-arguments.lua is turned into a module. All inside references to global variables are turned into formal parameters references. Now the l3build.lua main controller creates the global variable named Opts to replace the former variable that was created inside l3build-arguments.lua as a side effect.

I'm not sure about a name with a capital letter, plus I think `options` is a documented interface. Is there a deeper reason for renaming?

> All options["foo"] have been renamed to Opts.foo for a much better legibility.

Legibility is I suspect partly subjective: I prefer the `options["foo"]`, I guess partly a in TeX I'd have `\getoption{foo}` or similar. More objectively, the reason specifically for the format I went with here is that these are all the command line options. So the code emphasises that `options["foo"]` comes from `l3build <target> --foo`

I'll take a look at the detailed changes and make some comment, though as I say, a proper review is likely best by some other people (which I'll request).

l3build-arguments.lua
```diff
@@ -21,8 +21,8 @@ The development version of the bundle can be found at
 for those people who are interested.
 
 --]]
+local A = {}
```
josephwright, line 24: I'm not clear on this approach. I think you are using a local table to collect up various functions, but I'm not getting why that's superior to just adding local to those functions. I'm also wary of a very short table name being used other than in a very limited scope (e.g. inside one loop): seems a bit confusing to me.

I'd welcome comments from others!

l3build-arguments.lua
```diff
@@ -34,7 +34,7 @@ local insert           = table.insert
 
 -- Parse command line options
 
-option_list =
+A.option_list =
```
josephwright, line 37: Deosn't this make `option_list` purely local? How then do we add to it from a `build.lua` file?

l3build-arguments.lua
```diff
@@ -158,39 +158,38 @@ option_list =
 
 -- This is done as a function (rather than do ... end) as it allows early
 -- termination (break)
-local function argparse()
-  local result = { }
-  local names  = { }
-  local long_options =  { }
-  local short_options = { }
+A.argparse = function (arg)
+  local result = {}
```
josephwright, line 162: I take it you don't like the `{ }` convention :)

l3build-arguments.lua
```diff
@@ -158,39 +158,38 @@ option_list =
...
-  for k,v in pairs(option_list) do
-    if v["short"] then
-      short_options[v["short"]] = k
+  for k,v in pairs(A.option_list) do
+    if v.short then
+      short_options[v.short] = k
     end
     long_options[k] = k
   end
-  local args = args
```
josephwright, line 173: I have a feeling this was deliberate: I'll check back to be on the safe side

l3build-arguments.lua
```diff
@@ -277,31 +276,31 @@ local function argparse()
...
+    for _, engine in pairs(options.engine) do
       if not t[engine] then
         print("\n! Error: Engine \"" .. engine .. "\" not set up for testing!")
         print("\n  Valid values are:")
         for _, engine in ipairs(checkengines) do
           print("  - " .. engine)
         end
         print("")
-        exit(1)
+        os.exit(1)
```
josephwright, line 300: Is there a  reason for this change? Standard Lua guides say to use local copies of _everything_ in production code.

l3build.lua
```diff
@@ -115,45 +115,44 @@ unpackdir     = escapepath(unpackdir)
...
 --
 -- Deal with multiple configs for tests
 --
 
 -- When we have specific files to deal with, only use explicit configs
 -- (or just the std one)
-if options["names"] then
-  checkconfigs = options["config"] or {stdconfig}
+if Opts.names then
+  checkconfigs = Opts.config or {stdconfig} -- What is stdconfig?
```
josephwright, line 135: `stdconfig` is defined (like most other config data) in `l3build-variables`: it's the name of the main `build.lua` file, which of course could be something else ...

josephwright:
I've made a few comments but I'll leave it to others to review properly: as I wrote most of the code, that's only reasonable.

l3build-arguments.lua
zauguin: While I would keep the original code here for consistency reasons, it could be argued that specifically for `os.exit` not having a local copy is actually faster. (Since `os.exit` can never be called more than once, so the lookup is never duplicated)

l3build-aux.lua
```diff
@@ -68,11 +68,11 @@ end
 -- Do some subtarget for all modules in a bundle
 function call(dirs, target, opts)
   -- Turn the option table into a string
-  local opts = opts or options
+  local opts = opts or Opts
```
zauguin, line 71: This breaks compatibility with lots of existing code (including latex2e and latex3) and the use of uppercase variables violates l3build's coding conventions. Also why?

l3build-check.lua
```diff
@@ -55,7 +55,7 @@ local remove           = os.remove
 -- Set up the check system files: needed for checking one or more tests and
 -- for saving the test files
 function checkinit()
-  if not options["dirty"] then
+  if not Opts.dirty then
```
zauguin, line 58: While I would often be in favor of `a.b` instead of `a["b"]`, I would argue that for `options` in particular the `["..."]` should be kept even if the general convention is changed since quite often `["..."]` is required (because the key contains hyphens) and therefore switching between the syntax variants leads to inconsistent code.

l3build-install.lua
```diff
@@ -294,5 +294,5 @@ function install_files(target,full,dry_run)
 end
 
 function install()
-  return install_files(gethome(),options["full"],options["dry-run"])
+  return install_files(gethome(),Opts.full,Opts["dry-run"])
```
zauguin, line 297: This is the kind of inconsistency I mentioned in an earlier commit: The syntax suggests a difference between `full` and `dry-run` where no such difference exists.

jlaurens:
So sorry for being rude, I am very far from being a native English speaker. There was absolutely nothing personal, please consider my words to be purely technical.

I do not pretend that l3build needs a complete rewrite. I just see rather small places for big improvements.
Of course time is our enemy, and I do not pretend also that my pull requests must be merged in master right 
now.

> A few comments before getting to any code detail ;)
> 
[...]
> 
> > some of the globals should be initialized before use
> 
> I guess this might make more sense when I take a look at the code.

`os_time` at least.

> 
> > some globals should turn to local because they are only used in a very limited scope
> 
> Quite possibly, as long as they are not used outside @latexproject code.

Does "outside @latexproject code" mean documented in l3build.ltx?
Anyway, things like `ctan_field` should definitely be local.

> 
> > ctan_post should not be at the same time global and local.
> 
> Quite possibly: that one I'm not sure about.
> 
> > some locals are useless shallow copies of globals
> 
> I wonder if @blefloch is right here: are you talking _e.g._ `local print = print`?
> 
Not really. `l3build-aux.lua` reads
```
function call(dirs, target, opts)
  -- Turn the option table into a string
  local opts = opts or options
```
`opts` is a shallow copy of the eponym formal parameter. The same in `function ren(dir, source, dest)` at least.
BTW `opts` is unused, meaning that its value is always `options` whenever `call` is ... called.

`l3build-aux.lua` reads
```
  -- Keep data local
  local uploadconfig = uploadconfig
```
Despite being a local variable, further changes in `uploadconfig` are global.

Concerning benchmarks, I am far from convinced by the real benefits of making local copies here.
One evident thing is that the local copy must be used at least twice because it must compensate the supplemental allocation time. This is not always the case.

As example, the `table.insert` is often made local. The gain is about 15% for 10000 insertions, but that makes only 4 e-4 s. The gain is about 8% for 20 insertions, but that makes only 1 e-7 s. That does not ressemble a bottleneck when the overall process lasts several minutes.

In that respect, replacing `insert(t, x)` by `t[#t+1] = x` will improve the gain by a factor 2 which one would expect because appending is simpler than inserting.

There is a significant relative gain for the `print` function, but it is negligible here because the number or string displayed remains small. However, `local print = print` is absolutely harmless which is not the case of `local insert = table.insert`: 1) one has to remember that `insert` is just a shortcut of `table.insert`, 2) one cannot use `insert` for another purpose (or the other similar shortcuts). Of course these are not difficulties for the code writer himself, but these are for someone else.

> > The second goal will be to have some regression tests.
> 
> I'm not disagreeing, but just to note that the definitive question always is 'does a change to `l3build` show up when running `l3build ctan` in the LaTeX2e repo'.
> 

And other repositories as well.

As you mentioned the ultimate regression test is `l3build ctan` from the `latex` repository.
So changes must be made in a very safe and conservative way.
What would be nice is to keep `l3build` and have an "`l3build2`" to integrate new stuff.
Then the test would be to have `l3build ctan --debug` and `l3build2 ctan --debug` produce exactly the same `latex.zip`
And once `l3build2` is approved, it will replace `l3build`.

Moreover not only the `ctan` target can be used to compare old and new stuff.

> > In that pull request, l3build-arguments.lua is turned into a module. All inside references to global variables are turned into formal parameters references. Now the l3build.lua main controller creates the global variable named Opts to replace the former variable that was created inside l3build-arguments.lua as a side effect.
> 
> I'm not sure about a name with a capital letter, plus I think `options` is a documented interface. Is there a deeper reason for renaming?

`opts` is already used. This renaming allows

```
opts = opts or Opts
```
instead of
```
opts = opts or options
```
With that you can make a case insensitive search of `opts` and find "whatever" is related to options.

For the future, changing the name allows eventually to expose to `build.lua` a different `options` object which is a read only copy of `Opts`.

> 
> > All options["foo"] have been renamed to Opts.foo for a much better legibility.
> 
> Legibility is I suspect partly subjective: I prefer the `options["foo"]`, I guess partly a in TeX I'd have `\getoption{foo}` or similar. More objectively, the reason specifically for the format I went with here is that these are all the command line options. So the code emphasises that `options["foo"]` comes from `l3build <target> --foo`
> 
I am missing something. Technically
`options.foo` is closer to `l3build <target> --foo` than `options["foo"]`.
`options.foo` is also closer `\getoption{foo}` than `options["foo"]`
However, I agree that `\getoption{foo}` and `options["foo"]` both have delimiters but they do have a very different meaning.

> I'll take a look at the detailed changes and make some comment, though as I say, a proper review is likely best by some other people (which I'll request).

l3build.lua
```diff
@@ -28,31 +28,31 @@ for those people who are interested.
...
-build_kpse_path = match(lookup("l3build.lua"),"(.*[/])")
+local build_kpse_path = match(lookup("l3build.lua"),"(.*[/])")
 local function build_require(s)
-  require(lookup("l3build-"..s..".lua", { path = build_kpse_path } ) )
+  return require(lookup("l3build-"..s..".lua", { path = build_kpse_path } ) )
 end
 
 -- Minimal code to do basic checks
-build_require("arguments")
+A = build_require("arguments")
```
zauguin, line 52: Even if we would use `A` inside of `arguments`, we shouldn't use such a generic and short name for a global.

l3build.lua
```diff
@@ -28,31 +28,31 @@ for those people who are interested.
...
 local gsub             = string.gsub
-local next             = next
 local print            = print
-local select           = select
-local tonumber         = tonumber
 local exit             = os.exit
 
 -- l3build setup and functions
 kpse.set_program_name("kpsewhich")
-build_kpse_path = match(lookup("l3build.lua"),"(.*[/])")
```
zauguin, line 49: This again is breaking backward compatibility.

l3build.lua
```diff
@@ -115,45 +115,44 @@ unpackdir     = escapepath(unpackdir)
...
-        if options["halt-on-error"] then
+        if Opts["halt-on-error"] then
           exit(1)
         else
           insert(failed,checkconfigs[i])
         end
       end
     end
-    if next(failed) then
+    if #failed > 0 then
```
zauguin, line 155: I would expect `if failed[1]` to be even faster.

zauguin (changes requested):

l3build-arguments.lua
jlaurens: The name `A` simply reflects my lack of imagination. I am focusing on the principle of keeping local variables in one place. If variables must be shared, one function argument is sufficient. Avoiding globals is good practice.

l3build-arguments.lua
jlaurens: The problem is that a global `option_list` may be edited in `build.lua`. Is it what is expected ?
If so `option_list` should not be relied upon before `build.lua` is read.
The idea is to define a bunch of private variables making them local, then expose only some of them globally either RW or RO, and then import `build.lua`. Then we read only what we expect.

l3build-arguments.lua
jlaurens: Shorter, less eye movement, better reading...
Seriously, I wanted to discard the change but missed the step...

l3build-arguments.lua
jlaurens: `args` is never used. maybe a typo ?
`local arg = arg` ?
If the purpose is to make a local safe copy, then it better be at the top of the module.

l3build-arguments.lua
jlaurens: Eureka!
Now that makes more sense to me, this is not just a question of benchmark.
Then *everything* includes `print` and `checkengines` as well.
BTW, `check_engines` relies on `checkengines` and may be more appropriate somewhere else.

l3build.lua
jlaurens: I can't find any `stdconfig` out of `l3build.lua`.

l3build-arguments.lua
josephwright: The original syntax was `texlua build.lua` rather than `l3build`, which meant that `build.lua` was read _before_ `l3build.lua`. That's still supported by deprecated. As such, everything is (supposed to be) set up such that you can read `build.lua` either before or after `l3build.lua`. That's why all of `l3build.variables` is of the form `foo = foo or {<init>}`.

l3build-arguments.lua
josephwright: Probably a typo from some earlier refactoring, then: I was just checking as there are some oddities that are deliberate. I wanted to be sure, that was all.

l3build-arguments.lua
josephwright: > While I would keep the original code here for consistency reasons, it could be argued that specifically for `os.exit` not having a local copy is actually faster. (Since `os.exit` can never be called more than once, so the lookup is never duplicated)

Fair enough: I've not benchmarked, just went with a 'replace all of the globals' approach for standard stuff.

l3build.lua
josephwright: > I can't find any `stdconfig` out of `l3build.lua`.

Ah, right: this is one I think we renamed, so it could go yes.

l3build-arguments.lua
zauguin: > I am focusing on the principle of keeping local variables in one place.

That's generally something I can agree with, but it doesn't seem to be consistent with this PR making *more* internal functionality externally available. 


> If variables must be shared, one function argument is sufficient. Avoiding globals is good practice.

I don't see how function arguments relate to globals so I might be missing your point here. While avoiding globals (especially global names but to a smaller degree also effective globals like ones hidden in a table) is good practice, I would generally say the same about function arguments.

l3build-arguments.lua
zauguin: > If the purpose is to make a local safe copy, then it better be at the top of the module.

Well, if it was intended as performance enhancement, then it is better inside of the function if it is only used inside of the function and the function is only called once, since then it doesn't become a upvalue but just a regular stack entry. (Of course that mostly matters when it's accessed *a lot* which isn't the case here)

l3build-aux.lua
jlaurens: No need to break any `build.lua`. The global `options` seen in `build.lua` and the global `options` used by l3build need not be the same object, they just have to share some contents at some time. The question is: can `build.lua` change the `options` safely? 

Uppercasing the first letter is a convention used by linters. This is how we can make the difference between local and global variables. 

Of course this does not apply to `build.lua` which is completely independent.

Is there any other place where things could break ?

l3build-arguments.lua
zauguin: > Then _everything_ includes `print` and `checkengines` as well.

It should definitely include `print`. `checkengines` is different because we explicitly *do not* want it to be fixed, we actually expect it to change after the file is loaded (the global will normally be nil when the file is first loaded)

In an abstract way, the distinction is between "fixed functions" (like `exit`, `print`, `next`) on the one hand which both should never change and even if they would change we probably would still want to call the original variant, while on the other hand there is "data" (like `arg`, `checkengines`, etc, but might also include callbacks which might technically be functions) which is much more variable and therefore more thought has to go into creating local copies for these.

l3build-check.lua
jlaurens: The actual code is already inconsistent on that respect (see `option_list` definition)
This change is just consistent in the inconsistency...

l3build.lua
jlaurens: Can I have more precision ?
I cannot find any mention of this in the repository.

jlaurens:
Thanks for all these very interesting remarks.
I can see that an incremental change based on pull requests is complicated here.
Let me come back later with a more elaborate proposal.

l3build-aux.lua
zauguin: They don't have to be the same object, but separating the only to make them basically equivalent again seems kind of pointless.

Also yes, they can be changed safely (Exceptions: Changing `target` to `help` or `version` wouldn't work, but that would be kind of pointless anyway since build.lua can just call `help` or `version`) (Also yes, there are build.lua files using this feature.)

I'm a bit confused why a Lua linter would default to such settings since I'm not aware of any Lua libraries actually using that convention, but IMO a linter should be configured to be compatible with the project specific coding conventions and not the reverse, especially when the coding conventions affect documented external interfaces.

In which way is `build.lua` "completely independent"? It does share a bunch of global state with the rest of l3build, it is allowed to overwrite a lot of internal functions and is loaded without any isolation. That might not be how you would design such a system (it's certainly not how I would design such a system if I would start from scratch) but for `l3build` and any reasonably compatible system I would describe `build.lua` more as "closely coupled".

josephwright:
One thing I'd suggest, as a general approach, is to separate out PRs to be very focussed. For example, some of the 'style' changes would be fine, but I feel should go in across all files but as a single PR. For example, it seems the feel is for `foo = {}` rather than `foo = { }`, which could be done nicely in a single commit over all of `l3build`.

l3build-aux.lua
josephwright: > In which way is `build.lua` "completely independent"? It does share a bunch of global state with the rest of l3build, it is allowed to overwrite a lot of internal functions and is loaded without any isolation. That might not be how you would design such a system (it's certainly not how I would design such a system if I would start from scratch) but for `l3build` and any reasonably compatible system I would describe `build.lua` more as "closely coupled".

We started with quite a simple set up for just `latex3`, but as I've said expecting `texlua build.lua` as the lead-off. Supporting the LaTeX kernel requires quite a lot of customisation, and I didn't necessarily add it in the most ideal way initially. Some of that has been tided up (e.g. the format-building code hooks), but there are doubtless more places we could adjust.

One might argue that we could drop quite a lot of stuff: as a build system, stability isn't as important as it is for general code.

josephwright:
> Thanks for all these very interesting remarks.
> I can see that an incremental change based on pull requests is complicated here.
> Let me come back later with a more elaborate proposal.

BTW, please don't feel there is a sense of trying to put you off :) I'm very happy to see improvements, it's just that it's always best to talk them through first!

josephwright:
I was wondering if a style guide and some (matching) clean up would be useful? @jlaurens, @zauguin : thoughts?

PR #133, opened 2021-01-13
2) Truning l3build-help.lua into a module

jlaurens:
This pull request follows the one concerning l3build-arguments.lua. 

Removing any reference to a global variable inside l3build-help.lua.
- Some cosmetic changes to improve legibility
- `...:len()` changed to '#'
Actually  l3build-help.lua depends on l3build-stdmain.lua which is loaded afterwards. This is not a good code design and should be changed when l3build-stdmain.lua is changed.

### DIFF
--- a/l3build-help.lua
+++ b/l3build-help.lua
@@ -27,53 +27,55 @@ local match  = string.match
 local rep    = string.rep
 local sort   = table.sort
 
-function version()
-  print(
-    "\n" ..
-    "l3build: A testing and building system for LaTeX\n\n" ..
-    "Release " .. release_date .. "\n" ..
-    "Copyright (C) 2014-2020 The LaTeX3 Project"
-  )
+local H = {}
+
+H.version = function ()
+  print([[
+l3build: A testing and building system for LaTeX
+
+Release ]] .. release_date .. [[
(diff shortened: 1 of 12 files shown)


l3build-help.lua
```diff
@@ -82,10 +84,13 @@ function help()
...
-  print("")
-  print("Repository  : https://github.com/latex3/l3build")
-  print("Bug tracker : https://github.com/latex3/l3build/issues")
-  print("Copyright (C) 2014-2020 The LaTeX3 Project")
+  print([[
+Full manual available via 'texdoc l3build'.
+
+Repository  : https://github.com/latex3/l3build
+Bug tracker : https://github.com/latex3/l3build/issues
+Copyright (C) 2014-2020 The LaTeX3 Project
```
josephwright, line 92: We can't change this line without also adjusting the copyright-update code in the `build.lua` file for `l3build` itself (so tagging works properly). Same for the other copyight-containing lines.

josephwright:
General comments made in https://github.com/latex3/l3build/pull/132 apply here; as there, I think some review by at least one Lua expert would be good :)

jlaurens:
Let me come back later with a more convincing request.
Sorry for the disturbance.

PR #142, opened 2021-01-24
Refactor aux

jlaurens:
Minimize reference to global variables (turn epoch to an argument)
Some names are "more explicit"
snake_case cosmetic change.
Inline documentation added.

The @usage tag is used to mark functions as private/public. Public means "documented in l3build.dtx". Private means available to other l3build modules but not outside (not in build*.lua).

### DIFF
--- a/l3build-aux.lua
+++ b/l3build-aux.lua
@@ -103,7 +103,7 @@ function call(modules, target, opts)
   opts = opts or options
   local cli_opts = ""
   for k,v in pairs(opts) do
-    if k ~= "names" and k ~= "target" then -- Special cases, to be removed!!!
+    if k ~= "names" and k ~= "target" then -- Special cases, TODO enhance the design to remove the need for this comment
       local t = option_list[k] or {}
       local value = ""
       if t["type"] == "string" then
--- a/l3build-aux.lua
+++ b/l3build-aux.lua
@@ -22,6 +22,8 @@ for those people who are interested.
 
 --]]
 
+-- local safety guards and shortcuts
+
 local match = string.match
 
 local pairs = pairs
@@ -29,25 +31,41 @@ local print = print
 
 local lookup = kpse.lookup
 
+local os_time = os_time
 --
 -- Auxiliary functions which are used by more than one main function
 --
 
-function normalise_epoch()
+---Convert the given `epoch` to a number.
+---@param epoch string
+---@return number
+---@see l3build.lua
+---@usage private?
+function normalise_epoch(epoch)
+  assert(epoch, 'normalize_epoch argument must not be nil')
   -- If given as an ISO date, turn into an epoch number
   local y, m, d = match(epoch, "^(%d%d%d%d)-(%d%d)-(%d%d)$")
   if y then
-    epoch =
-      os_time({year = y, month = m, day = d, hour = 0, sec = 0, isdst = nil}) -
-      os_time({year = 1970, month = 1, day = 1, hour = 0, sec = 0, isdst = nil})
+    return os_time({
+        year = y, month = m, day   = d,
+        hour = 0, sec = 0, isdst = nil
+      }) - os_time({
+        year = 1970, month = 1, day = 1,
+        hour = 0, sec = 0, isdst = nil
+      })
   elseif match(epoch, "^%d+$") then
-    epoch = tonumber(epoch)
+    return tonumber(epoch)
   else
-    epoch = 0
+    return 0
   end
 end
 
-function setepoch()
+---CLI command to set the epoch, will be run while checking or typesetting
+---@param epoch string
+---@return string
+---@see check, typesetting
+---@usage private?
+function set_epoch_cmd(epoch)
   return
     os_setenv .. " SOURCE_DATE_EPOCH=" .. epoch
       .. os_concat ..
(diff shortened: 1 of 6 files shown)


davidcarlisle:
Before any PR can be considered we would have to confirm that it has no effect on (at least) the latex3 and latex2e test suites. Can you confirm that these changes do not affect the builds in those repositories?

josephwright:
Seems to largely run at the top level up to `l3benchmark`, where I got an odd error. Trying to investigate, I then found
```
palladium l3benchmark % l3build check         
/Users/joseph/Library/texmf/scripts/l3build/l3build-aux.lua:45: normalize_epoch argument must not be nil
```
and got the same for other single folders. So something's not quite right.

josephwright:
Oh, ignore that: was my fault. Let me track down the original oddity.

josephwright:
OK, there is definitely something not quite right. If you use the PR version and run it for `l3benchmark`, you get a failure as the lines
```
Lua function __sys_elapsedtime: = 11
Lua function __sys_resettimer: = 12
```
are not normalised to
```
Lua function __sys_elapsedtime: = ...
Lua function __sys_resettimer: = ...
```
I don't see immediately why this would be the case: there's no change that looks suspicious. But something must be happening: it's reproducible and with my system `l3build` all is fine. (`l3benchmark` can be a bit fussy.)

l3build-aux.lua
```diff
@@ -57,69 +75,89 @@ function setepoch()
...
-    if errorlevel ~= 0 then
-      return errorlevel
+---Unpack the given dependencies.
+---A dependency is the path of a directory relative to the main one.
+---@param deps table regular array of dependencies.
+---@return number 0 on proper termination, a non 0 error code otherwise.
+---@see stdmain, check, unpack, typesetting
+---@usage Private?
+function dep_install(deps)
+  local error_level
```
josephwright, line 155: I guess other renames are a matter of 'style' (I favoured single runs of `a-z` when I wrote `l3build`, others have added `_`), but for `errorlevel`, there is a long-established convention that it's one word.

l3build-aux.lua
```diff
@@ -57,69 +75,89 @@ function setepoch()
...
+---@usage Public
+function call(modules, target, opts)
+  -- Turn the option table into a CLI option string
+  opts = opts or options
+  local cli_opts = ""
   for k,v in pairs(opts) do
-    if k ~= "names" and k ~= "target" then -- Special cases
-      local t = option_list[k] or { }
-      local arg = ""
+    if k ~= "names" and k ~= "target" then -- Special cases, to be removed!!!
```
josephwright, line 106: I'm not sure why you'd want to remove these

josephwright:
Oh, mumble: ignore me! There is a change on `master` that's not yet been released: I'd entirely forgotten about it. So the PR is fine in that sense: we do have a pending change required to `.tlg` files, that's all.

I'll wait to see what @zauguin  and @cereda think of the detail of the changes

l3build-aux.lua
jlaurens: This is a matter of both style and efficiency. `errorlevel` smells a bit vintage TeX isn't it? Nowadays we have `\NewDocumentCommand` and `\l_tmpa_str` ;-)
BTW, is it told somewhere that the error level must be nonnegative?

l3build-aux.lua
jlaurens: Not a remove.
`arg` has been renamed to `value` because `arg` is already used by Lua, it may be misleading.

l3build-aux.lua
josephwright: All I mean is that `errorlevel` is a very widely used name (e.g. Windows batch files), so I'd favour sticking to it.

In terms of possible return values, I have 'fun' with tools we call setting odd numbers, I think due to twos-complement somewhere. So the only thing I would rely on is `0` is 'success' and non-zero is 'failure'.

l3build-aux.lua
josephwright: You added `to be removed!!!`, which is what I didn't follow. 

On using `arg`, this is local so it shouldn't be a big issue (we don't need the CLI `arg`  table here), but I'm not opposed to a rename.

l3build-aux.lua
```diff
@@ -57,69 +75,89 @@ function setepoch()
...
-    for _,v in pairs(opts["names"]) do
-      s = s .. " " .. v
+  if opts.names then
+    for _, name in pairs(opts.names) do
+      cli_opts = cli_opts .. " " .. name
     end
   end
-  local scriptname = getscriptname()
-  for _,i in ipairs(dirs) do
-    local text = " for module " .. i
```
josephwright, line 99: I don't see where this text gets moved to ...

josephwright:
> Before any PR can be considered we would have to confirm that it has no effect on (at least) the latex3 and latex2e test suites. Can you confirm that these changes do not affect the builds in those repositories?

For the avoidance of doubt: I have checked this is the case

l3build-aux.lua
jlaurens: The fact that `errorlevel` is already known in a widely used different context and a different usage also justifies to use a different name here. 

This is not a 0/non-zero alternative, it is a 0/positive alternative. clean, typeset and install both cumulate the error levels of various commands. If exactly one command returned 1 whereas exactly another returned -1, then the sum would be 0 and the result would be a falsy success.

josephwright: > The fact that `errorlevel` is already known in a widely used different context and a different usage also justifies to use a different name here.
> 
> This is not a 0/non-zero alternative, it is a 0/positive alternative. clean, typeset and install both cumulate the error levels of various commands. If exactly one command returned 1 whereas exactly another returned -1, then the sum would be 0 and the result would be a falsy success.

To-date, we've not had a case where that's an issue. I'd suggest tighening up where values return from an external process: we can do something like `if os.execute("thing") > 0 then errorlevel = errorlevel + 1` or similar. To be honest, we could also just fail at the first dodgy return: in the end, that's likely better style.

josephwright: So `if os.execute("thing") then return 1 end` for all `thing` might be best.

l3build-aux.lua
jlaurens: one line 134

l3build-aux.lua
jlaurens: Oups!!! `target` and `names` contradict the fact that `options` keys are CLI options. Replacing them by global variables or replacing the keys by "target\0" and "names\0" would resolve that contradiction.

The rename is a detail but details really make life easier. If we want to know every place where the global `arg` is used, then no need to manually filter out a find output.

josephwright: I guess this is a matter of how one things about it: I see `options` as 'everything after `l3build`' on the CLI. For example, there is no certainty that the 'names' really are names, they could be some other optional data (OK, then perhaps better called 'final list' or similar).

josephwright: That said, I have no issue with a change _provided_ it's documented: we do currently say that everything, including the target and the list of names, is in that table.

cereda:
Two out of the three commits associated to this PR have the same title. Although their associated messages are a bit more descriptive, I'd highly suggest to avoid doing this in the upcoming commits.

cereda (approved):
Changes are compliant. @josephwright Could you confirm there's no regression?

josephwright:
> Changes are compliant. @josephwright Could you confirm there's no regression?

Yes, it's all OK for the core tasks

PR #144, opened 2021-01-27
global #1: new global module and first unit tests

jlaurens:
No modification made to current files.

`l3build-global.lua` is added as a starting point to collect some globals into one namespace.

`testdir/` folder is added to collect the lua testing material. The README.md may help.

The "#1" indicates that there will be more steps (much more steps actually) it might be advisable not to merge in master but in a `dev` branch. I would prefer a `future` branch to avoid the pressure of a TL deadline.

Next, this will be plugged into `l3build`, one step at a time.

Comments and suggestions are welcome. Explanations may seem sometimes very light.

BTW, we can use a "minifier" to strip off comments before installation.
Maybe a "uglifier" also.

### DIFF
--- a/l3build-global.lua
+++ b/l3build-global.lua
@@ -1,0 +1,91 @@
+--[[
+
+File l3build-global.lua Copyright (C) 2018-2020 The LaTeX3 Project
+
+It may be distributed and/or modified under the conditions of the
+LaTeX Project Public License (LPPL), either version 1.3c of this
+license or (at your option) any later version.  The latest version
+of this license is in the file
+
+   http://www.latex-project.org/lppl.txt
+
+This file is part of the "l3build bundle" (The Work in LPPL)
+and all files in that bundle must be distributed together.
+
+-----------------------------------------------------------------------
+
+The development version of the bundle can be found at
+
+   https://github.com/latex3/l3build
+
+for those people who are interested.
+
+--]]
+
+---@module global
+---For global data and methods,
+---Very first to be loaded
+---@usage kpse.set_program_name("kpsewhich"); local l3b = require(kpse.lookup("l3build-global.lua", { path = kpse.lookup("l3build.lua"):match("(.*[/])") }))
+
+
+-- Local safe guards and shortcuts
+
+local kpse = kpse
+local lookup  = kpse.lookup
+
+-- l3build setup and functions
+
+-- next is like `package.path`. Must end with  `;`
+local package_path = ""
+
+---`<dir_1>`, `<dir_2>`,... are the paths of supplemental directories where
+---`l3build*.lua` files should be looked for.
+---Names need not end with a `/`.
+---@param dir1 string
+---@param dir2 string
+---@usage `l3b.package_dir_append(testdir)`
+local function package_dir_append(...)
+  local arg = table.pack(...)
+  for i = 1, arg.n do
+    local arg_i = arg[i]
+    if #arg_i > 0 then
+      package_path = package_path
+      .. arg_i .. "/l3build-?.lua;"
+      -- .. arg_i .. "/l3build-?/init.lua;" -- for directory packages
+    end
+  end
+end
+
+local loaded = {} -- cache the result of require queries
+
+kpse.set_program_name("kpsewhich")
+local build_kpse_path = lookup("l3build.lua"):match("(.*[/])")
+
+---Require `l3build-<key>.lua`.
+---Can be used multiple times for the same `<key>` from different files.
+---@see package_dir_append to add lookup directories
+---@param key string
+---@return table or nil
+local function l3b_require(key)
+  local was = loaded[key]
+  if not was then
+    local path = lookup("l3build-" .. key .. ".lua", { path = build_kpse_path })
+    if not path then
+      path = package.searchpath(key, package_path) -- useful for testing
+      if not path then
+        error("Unsupported l3build package key: " .. key)
+      end
+    end
+    was = { ans = dofile(path) }
+    loaded[key] = was
+  end
+  return was.ans
+end
+
+return {
+  _TYPE = 'module',
+  _NAME = 'global',
+  _VERSION = '2021-01-26',
+  require = l3b_require,
+  package_dir_append = package_dir_append,
+}
--- a/testdir/README.md
+++ b/testdir/README.md
@@ -1,0 +1,31 @@
+# Testing
+
+testing lua scripts is performed with the aid of [LuaUnit](https://luaunit.readthedocs.io/en/luaunit_v3_2_1/)
+
+All test material related to lua is gathered in the `testdir` folder.
+
+## Example
+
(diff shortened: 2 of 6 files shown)


davidcarlisle:
Having unit tests is reasonable but adding luaunit is almost doubling the amount of Lua in this repository and much of the luaunit framework is either not needed (we have almost no numeric code so assertions against NaN etc not that helpful) and much of the rest is a test harness which overlaps the functionality of l3build itself.

Wouldn't a more consistent approach be to allow lua (rather than tex) test files but use the l3build infrastructure to marshal the tests and report results rather than importing an entirely different test framework?  It seems to me that a new lua test file possibility plus a basic library of lua assertions matching the tex tests we have in regression-test.tex  would allow testing of the lua code here and also usefully be used with other tex related projects that have a lua component.

jlaurens:
Adding lua testing capabilities to l3build might be very interesting indeed, but it is not really the purpose of this PR.

It happens that unit testing has more to offer than l3build concerning test management.
For example test suites are very practical during development.

Also, numeric assertions may be useful while creating documents with dynamic pseudo random values.

davidcarlisle:
Yes I realise that that is not the focus of the PR but that is rather the point of my comment. I'll let others comment but personally I think that checking in an entirely different test harness source into the sources of the l3build test system is a very unnatural development and distorts the balance of the sources here.  If there are specific tests (including Lua assertions) that we need we could add them to l3build without  adding over three thousand lines of luaunit sources here.

FrankMittelbach:
> Yes I realise that that is not the focus of the PR but that is rather the point of my comment. I'll let others comment but personally I think that checking in an entirely different test harness source into the sources of the l3build test system is a very unnatural development and distorts the balance of the sources here. If there are specific tests (including numeric Lua assertions) that we need we could add them to l3build without adding over three thousand lines of luaunit sources here.

agreed. We should keep this lean.

jlaurens:
Nothing is checked into the sources. This is the absolute beauty of this PR.
All the testing material is gathered into only one directory.
You can simply throw away this directory whenever you want for whatever reason you want
and everything will be like if it never existed.
Being orthogonal to the sources, there is no reason why this PR would prevent to ever add any lua testing capabilities to l3build, on the contrary I would say.

davidcarlisle:
It is in the repository which is what I meant by the sources, and it just seems an odd choice to use an entirely different test system when testing is the point of l3build.

That isn't a criticism of luaunit in general, it just seems inappropriate choice here.

jlaurens:
How long should I wait before l3build will have lua testing facilities ?

davidcarlisle:
> 
> 
> How long should I wait before l3build will have lua testing facilities ?

You are suggesting adding them, I just commented that adding them to the existing testing framework would be more consistent than using a completely different framework.

josephwright:
On the 'testing Lua itself' aspect, @zauguin started a PR a while ago that broadens testing (https://github.com/latex3/l3build/pull/128). That one got a bit stalled, but I think if we round it out it likely can help here.

I'm not sure about a merge to `master`: this is clearly part of 'a bigger effort' but at the moment feels like it would be best as a longer-term branch. Something like `self-testing` or `modularisation`?

The bigger question I've got is does it really help moving the interfaces from where they are defined (in the 'relevant' `l3build-....lua` files) to a single file? I'm not sure what the 'usual' is: my other Lua experience is basically ConTeXt, and it doesn't I think work that way.

jlaurens:
@davidcarlisle 
Sorry for being unclear, the purpose of this PR and the next ones is not to add any lua testing capabilities to `l3build`. When I said that next step would be to plug this into l3build I really meant to plug only `l3build-global.lua`.

davidcarlisle:
But still I don't think the PR should be merged while it is adding luaunit.

More generally you seem to be pushing for a refactoring of l3build for only vaguely specified benefits of improving Lua idioms. 

I just checked and I have 285 build.lua files here for assorted ctan packages to which I have write access.  The best that can happen from any re-write is that you don't break these builds. There is no obvious sign that you are testing the changes with a range of existing builds (at the very least the latex2e and latex3 ones, but also hyperref and the ho-tex packages.)
If you suggest any changes then I think at minimum you should confirm that the latex3 and latex2e build.lua are not affected. That is_far_ more important testing for any l3build change than any unit tests that you might add.

zauguin:
> `l3build-global.lua` is added as a starting point to collect some globals into one namespace.

Any attempt to "collect globals into one namespace" would either end up making this namespace part of the global namespace later on and therefore render it useless, or would have to break backward compatibility. I am not convinced that any of these changes are sufficiently motivated to break compatibility on this scale.

> 
> `testdir/` folder is added to collect the lua testing material. The README.md may help.
> 
> The "#1" indicates that there will be more steps (much more steps actually) it might be advisable not to merge in master but in a `dev` branch. I would prefer a `future` branch to avoid the pressure of a TL deadline.

As I said above, I am not convinced that we want this merged even in a longer term, but if we would I would always prefer branches wit names telling something about their content over generic names like `future`.


> BTW, we can use a "minifier" to strip off comments before installation.
> Maybe a "uglifier" also.

We could, but why would we want that? l3build is designed in a way where most of it's internals can be adjusted to individual use cases, so having a commented source version available on every installation is an important feature.

cereda:
> BTW, we can use a "minifier" to strip off comments before installation.
> Maybe a "uglifier" also.

I agree with @zauguin, there is no need for them. Minifiers are typically used to reduce transfer time over a network and not for speeding up parsing or execution times. `l3build`'s appeal is to be customised _ad hoc_ to certain scenarios, so a good, structured, easy to read code and comments are needed. Besides, running such tools in a release chain adds yet another overhead which provides no practical value at all. The code has to be understandable.

Consider these two test files:

| File | Lines | Bytes | Chars | Words | Comments |
|:---|---:|---:|---:|---:|---:|
| `script1.lua` | 1394 | 13381 | 94696 | 94696 | 1393 |
| `script2.lua` |1 | 2 | 21 | 21 | 0 |

Running a benchmark tool with a warmup of 100 runs and a minimum 100 executions, this is the result I got:

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `lua script1.lua` | 1.6 ± 0.2 | 1.0 | 2.5 | 1.27 ± 0.20 |
| `lua script2.lua` | 1.2 ± 0.2 | 0.7 | 1.9 | 1.00 |

which is IMHO pretty negligible. As long as no one plans to include their own _War and Peace_ inside the code, there is absolute no need for any minifier. :wink: Besides, the _real_ bottleneck of `l3build` lives outside the tool itself, as it only invokes external commands and acts accordingly based on their output. So even with the most honest benchmark, it would be quite unfair to blame the existing code alone to justify the use of minifiers.

As for uglifiers, readability and usage reasons previously state apply.

<details>
  <summary>Spoilers</summary>

  ooh `xii.lua`
</details>

jlaurens:
Aknowledge

PR #146, opened 2021-01-28
Alternate to PR #145 Document `options.target` and reserve keywords

jlaurens:
Document `options.target` in `l3build.dtx`.
The table description page 28 was not consistent is the use of capitals. As lua types are lowercase, all types are changed to lowercase.

Add a `reserved` tag to option definitions in `option_list` to reserve keywords.
Ideally, optional_list should be an implementation detail.

### DIFF
--- a/l3build-arguments.lua
+++ b/l3build-arguments.lua
@@ -154,14 +154,14 @@ option_list =
         desc = "Print version information and exit",
         type = "boolean"
       },
-      target = 
-        {
-          reserved = true,
-        },
-      names = 
-        {
-          reserved = true,
-        },
+    target = 
+      {
+        reserved = true,
+      },
+    names = 
+      {
+        reserved = true,
+      },
   }
 
 -- This is done as a function (rather than do ... end) as it allows early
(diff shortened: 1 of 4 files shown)


josephwright:
> The table description page 28 was not consistent is the use of capitals. As lua types are lowercase, all types are changed to lowercase.

My aim was they were descriptors, but it's not I guess super-important. (Personally, I dislike Boolean being given with a lower case initial, as it's derived from a proper noun.) Probably it they are meant to match the Lua types then they should be in `\texttt`.

josephwright:
I'd welcome other people's input here: I guess I favour this approach over moving the data into multiple tables. Yes, I do see that the `names` and `target` are not _options_, but I don't think I'd ever fancy
```
l3build some-target --names=foo,bar --target=baz oh-no-another-name
```
so their 'loss' is not a major problem.

jlaurens:
In PiL online at least, types are in lowercase italic, no tt.
Moreover this table can read like
"target is a string"
"debug is a boolean"
"engine is a table"
In the whole document, the type is always typeset in lowercase roman.

jlaurens:
If we happen to allow people to customize the options list like they can customize the target list, it will suffice to mention that one can only add keys that are not already in this table.

PR #147, opened 2021-01-28
Useless line removed

jlaurens:


josephwright:
I'm not sure why you feel this is redundant

jlaurens:
unused variable

u-fischer:
resultdir is used l3build-check.lua, it is the path where the real pdf's are stored. I need them to check their validity.

jlaurens:
line 170/169

jlaurens:
Lines up to 170 read
```
    if next(failed) then
      for _,config in ipairs(failed) do
        print("Failed tests for configuration " .. config .. ":")
         print("\n  Check failed with difference files")
         local testdir = testdir
         if config ~= "build" then
           resultdir = resultdir .. "-" .. config ------------- resultdir is unused simply...
           testdir = testdir .. "-" .. config
         end
         for _,i in ipairs(filelist(testdir,"*" .. os_diffext)) do
          print("  - " .. testdir .. "/" .. i)
        end
        print("")
      end
      exit(1) ---------------------------------------------- because the program ends here
    else
```
BTW, the value of resultdir at the end of the loop is `<resultdir>-<failed config_1>-<failed config_2>-...-<failed config_n>`. I guess it is not really expected. A second local declaration is certainly missing.
In fact, someone has been bitten by the "all is global" policy, fortunately it was harmless, for this time at least.

josephwright:
> In fact, someone has been bitten by the "all is global" policy, fortunately it was harmless, for this time at least.

The multi-configuration stuff is in some places a bit rough, as it was added well after the main code was written. However, `resultdir` is a configuration setting so has to be global (and relates to a directory in the file system). I don't think me overlooking the exact set up here is a reason to alter the entire approach!

jlaurens:
No, but this approach must be taken more cautiously.
LaTeX3 has a de facto "all is global" policy but also has the "g" and "l" prefix.
That makes a real difference.

PR #158, opened 2021-02-03
some error management abspath

jlaurens:
Due to the use of `lfs.chdir`, the return value is not the expected one
- when the target directory does not exist
- when permission were not sufficient
In both situations, lfs.currentdir() returns a path unrelated to the given argument.
These restrictions are added to the dtx.
Now the abspath returns what chdir returns in case of error.

### DIFF
--- a/l3build-file-functions.lua
+++ b/l3build-file-functions.lua
@@ -171,12 +171,16 @@ function normalize_path(path)
 end
 
 -- Return an absolute path from a relative one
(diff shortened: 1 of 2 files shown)


josephwright:
Should we just hard-error (`os.exit`) under these circumstances, as the return value is still problematic

jlaurens:
New commit with an error() instead of the return.
This exits with a message except when pcall'ed

PR #160, opened 2021-02-06
Use LPEG to convert glob expression into pattern

zauguin:
Reimplement `glob_to_pattern` using LPEG to make it more extensible and (IMO) more readable.

The resulting pattern is still a Lua pattern and not a LPEG parser. This ensures that the function keeps compatibility and it is faster for typical file globs.

### DIFF
--- a/l3build-file-functions.lua
+++ b/l3build-file-functions.lua
@@ -49,75 +49,22 @@ local gsub             = string.gsub
 
 local insert           = table.insert
 
--- Convert a file glob into a pattern for use by e.g. string.gub
--- Based on https://github.com/davidm/lua-glob-pattern
--- Simplified substantially: "[...]" syntax not supported as is not
--- required by the file patterns used by the team. Also note style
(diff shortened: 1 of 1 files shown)


jlaurens:
What a coincidence, I had rewritten the `glob_to_pattern` to fix bugs, and add support for "**" and classes, but I was wondering why not using `lpeg` as it is available in `texlua`.

josephwright:
> What a coincidence, I had rewritten the `glob_to_pattern` to fix bugs, and add support for "**" and classes, but I was wondering why not using `lpeg` as it is available in `texlua`.

'I don't understand LPEG' would cover it nicely: to me, patterns are _much_ clearer. But I'm happy if some experts wish to do other stuff. (I'm intrigued about issues with `glob_to_pattern`, which as the source says is just a re-formatted copy of code I used under the MIT license.)

josephwright:
This doesn't pass: is it still 'live'?

PR #162, opened 2021-02-08
issue#161: Solved

jlaurens:
1) file-functions:
a) new `copy` function for only one item
b) `cp` now uses `copy`
2) `copy` is documented in the `dtx`
3) For the other files, every `cp` has been revisited
a) replaced by `copy` for item names
b) var names are replaced to include a `g` when it is a glob to avoid confusion.

### DIFF
--- a/l3build-check.lua
+++ b/l3build-check.lua
@@ -63,23 +63,23 @@ function checkinit()
   -- Copy dependencies to the test directory itself: this makes the paths
   -- a lot easier to manage, and is important for dealing with the log and
   -- with file input/output tests
-  for _,i in ipairs(filelist(localdir)) do
-    cp(i, localdir, testdir)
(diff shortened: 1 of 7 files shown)


wspr:
Hmm. It's late here and my brain is fuzzy, and reading over the new code really made me have to think about each case whether it should be `cp` or `copy`. Is having a loop that executes only once such a big deal here? I think it complicates the code (two commands instead of one) and may introduce subtle bugs over time if something that starts as `copy()` needs to become `cp()` for some reason. So I would probably not merge this one -- my 2¢ only though.

If this approach were to be retained, I'd strongly suggest changing the names to something like `copy()` and `copy_file()` to give them more direct meaning if read out of context. Very little chance if I stumbled over the code one day whether I would remember which of `copy()` or `cp()` was the one that worked on globs.

PR #163, opened 2021-02-08
glob_to_pattern_x

jlaurens:
This is an example for discussion mainly.
The benefits of using LPEG in PR #160 do not seem evident at first glance.
I have the feeling that using LPEG in my glob_to_pattern_x might be the way to go
if someone with a greater experience would take a look.

This is the original glob_to_pattern (not simplified) with enhancements following
https://www.man7.org/linux/man-pages/man7/glob.7.html:
1) "*" does not pass through directory boundaries
2) new "**" passes through directory boundaries
3) support for character ranges like [a-z], always excuding '/'
4) support for classes like [:letter:]
5) No utf* support of course

### DIFF
--- a/l3build-file-functions.lua
+++ b/l3build-file-functions.lua
@@ -49,6 +49,8 @@ local gsub             = string.gsub
 
 local insert           = table.insert
 
+local byte             = string.byte
+
 -- Convert a file glob into a pattern for use by e.g. string.gub
 -- Based on https://github.com/davidm/lua-glob-pattern
 -- Simplified substantially: "[...]" syntax not supported as is not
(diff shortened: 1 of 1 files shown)


josephwright:
>     3. support for character ranges like [a-z], always excuding '/'
> 
>     4. support for classes like [:letter:]


They were deliberately excluded as they don't work at the Command Prompt.

jlaurens:
Can you explain how glob_to_pattern is related to the command prompt/shell?

josephwright:
Because it's consistent: 

> Could you explain how glob_to_pattern is related to the command prompt/shell?

Largely a question of consistency: you want the same matching in the Lua as you've have if you typed things in at the Terminal/Command Prompt, as far as possible.

FrankMittelbach:
> Largely a question of consistency: you want the same matching in the Lua as you've have if you typed things in at the Terminal/Command Prompt, as far as possible.

the point here being the normal (expected) user of l3build is a TeX programmer not a lua programmer

josephwright:
More seriously, the aim was to cover the simple case of `*.ext` or `foo????.ext`, rather than anything more complex.

jlaurens:
What exactly prevented the original glob_to_pattern to cover those simple cases ?

josephwright:
> What exactly prevented the original glob_to_pattern to cover those simple cases ?

I'm lost: it does

jlaurens:
Then why limiting the features?

josephwright:
> Then why limiting the features?

The aim was to have the feature we needed without going over-the-top. The idea with the glob was that it is useful: install 'whatever `.sty` files or whatever. Beyond that, we've only needed hard-coded lists.

jlaurens:
That made some sense.
My question is motivated from a user POV who only reads the `dtx`.
Then limitations do not find real explanations and that is somehow frustrating.

josephwright:
> That made some sense.
> My question is motivated from a user POV who only reads the `dtx`.
> Then limitations do not find real explanations and that is somehow frustrating.

I guess I never saw this as a 'limitation': file specs are always given as `<name>.<ext>` or involve `*` and/or`?`, that's just how they work. So `l3build` is being consistent with everything else.

jlaurens:
Again this can make sense within the latex3 GitHub repository. But when reading the doc, nothing like that is explained. The doc mention globs, even "**" globs which is an advanced feature, but does not say what is not allowed with that globs compared to the standards.

PR #164, opened 2021-02-08
no hidden files in filelist by default

jlaurens:
Move the declaration of pattern inside its scope
Exclude dot names from default listing

there is often a `.DS_Store` in a directory on OSX, maybe a `.git` too

### DIFF
--- a/l3build-file-functions.lua
+++ b/l3build-file-functions.lua
@@ -278,9 +278,14 @@ function filelist(path, glob)
           insert(files, entry)
         end
       else
-        if not match(entry, "^%.") then
-          insert(files, entry)
-        end
+        insert(files, entry)
+      end
+    end
+    local t = files
+    files = {}
+    for _, entry in pairs(t) do
+      if not match(entry, "^%.") then
+        insert(files, entry)
       end
     end
   end
--- a/l3build-file-functions.lua
+++ b/l3build-file-functions.lua
@@ -267,18 +267,18 @@ end
 -- if absent
 function filelist(path, glob)
   local files = { }
-  local pattern
-  if glob then
-    pattern = glob_to_pattern(glob)
-  end
   if direxists(path) then
+    local pattern
+    if glob then
+      pattern = glob_to_pattern(glob)
+    end
     for entry in lfs_dir(path) do
       if pattern then
         if match(entry, pattern) then
           insert(files, entry)
         end
       else
-        if entry ~= "." and entry ~= ".." then
+        if not match(entry, "^%.") then
           insert(files, entry)
         end
       end
--- a/l3build-file-functions.lua
+++ b/l3build-file-functions.lua
@@ -281,11 +281,15 @@ function filelist(path, glob)
         insert(files, entry)
       end
     end
-    local t = files
-    files = {}
-    for _, entry in pairs(t) do
-      if not match(entry, "^%.") then
-        insert(files, entry)
+    if not pattern
+    or pattern:gsub("%%", ""):gsub("%*", ""):match("%*")
+    then
+      local t = files
+      files = {}
+      for _, entry in pairs(t) do
+        if not match(entry, "^%.") then
+          insert(files, entry)
+        end
       end
     end
   end
--- a/l3build-file-functions.lua
+++ b/l3build-file-functions.lua
@@ -266,6 +266,7 @@ end
 -- Generate a table containing all file names of the given glob or all files
 -- if absent
 function filelist(path, glob)
+  print("************************************* FILELIST")
   local files = { }
   if direxists(path) then
     local pattern
@@ -281,13 +282,16 @@ function filelist(path, glob)
         insert(files, entry)
       end
     end
-    if not pattern
-    or pattern:gsub("%%", ""):gsub("%*", ""):match("%*")
-    then
+    if not pattern or not match(pattern, "^%^%%%.") then
       local t = files
       files = {}
-      for _, entry in pairs(t) do
-        if not match(entry, "^%.") then
+      for _, entry in ipairs(t) do
+        if match(entry, "^%.") then
+          print("**** ignored " .. entry)
+          if entry == ".tex" then
+            print(".tex is ignored during " .. options.target)
+          end
+        else
           insert(files, entry)
         end
       end


josephwright:
Will this still allow `.` files when necessary? We need to allow for at least `.tex`.

josephwright:
To be clear, we have to be able to install dot files: in LaTeX2e we have a file called `.tex` that is vital.

jlaurens:
I see, but one would not want to install 3rd party dot file.
In the new commit, the dot files are filtered out only if the pattern contains a star.
I suspect that ".tex" files should be copied with a copy file method, not a copy blog method.

josephwright:
> I see, but one would not want to install 3rd party dot file.

Well true, but that's the general question of the balance between the convenience of globs and the accuracy of an explicit list of single file names.

> In the new commit, the dot files are filtered out only if the pattern contains a star.
> I suspect that ".tex" files should be copied with a copy file method, not a copy glob method.

The issue there is we have a *load* of `.tex` files, one of which is `.tex`: it' not a hidden file, it's a file with no 'name' part. [Remember that dot file are really only hidden by accident, in contrast to on Windows where hidden is a proper feature :)]

davidcarlisle:
On Mon, 8 Feb 2021 at 17:08, LAURENS Jérôme <notifications@github.com>
wrote:

> I see, but one would not want to install 3rd party dot file.
> In the new commit, the dot files are filtered out only if the pattern
> contains a star.
>


This would break the latex2e/tools/build.lua

installfiles = {"*.def", "*.sty", "*.tex"}

matches .tex

Please, a minimum requirement before making any pull request is to confirm
that it does not break the latex2e release.


l3build ctan

at the top level in the latex3/latex2e git repository has to work.

David



I suspect that ".tex" files should be copied with a copy file method, not a

FrankMittelbach:
> I see, but one would not want to install 3rd party dot file.

what is your use case here? maliceous files? or just the mentioned `.git` `.DS_Store` and the like?
The former you can't prevent on a script based system and the latter should pose no problem unless you do `*` copies without control over what is in the directory

> In the new commit, the dot files are filtered out only if the pattern contains a star.
> I suspect that ".tex" files should be copied with a copy file method, not a copy glob method.

well, no, because it is generated during "unpack" and then moved from one directy to others, e.g. the test directories. So there are specifications like `"*.tex", "*.sty", ....` which also (deliberately) cover the file `.tex`

jlaurens:
.DS_Store files are system files automagically generated by OSX. We have absolutely no control about it.

I could not make full tests on latex2e because the latex2e rep si actually broken:
```
Failed tests for configuration config-TU:

  Check failed with difference files
  - ../build/test-config-TU/github-0332.xetex.diff
```
When I wanted to diff some zip files, I was not even able to create any.
So my tests in fixing the dot files problem are focused on the install part.

1) my "expected" is the local texmf/tex folder populated with TL 2020 l3build and GitHub latex2e.
2) my "actual_1" is the local texmf/tex folder populated with TL 2020 l3build and the modified GitHub latex2e
3) my "actual_2" is the local texmf/tex folder populated with the modified l3build and the modified GitHub latex2e

All 3 folders show absolutely no diff except the new `regression-test.tex` of course.

Here is the modification on the latex2e side: a dedicated ".tex" glob is added.
- in `./required/tools/build.lua`
```
installfiles = {"*.def", "*.sty", "*.tex", ".tex"}
```
On the l3build side, the change is in the last commit.
The idea is to only accept dot files when the glob does require it (just like standard globs).

josephwright:
> .DS_Store files are system files automagically generated by OSX. We have absolutely no control about it.

Indeed (or almost: I believe it can be disabled, though that's probably more work than it's worth).

> I could not make full tests on latex2e because the latex2e rep si actually broken:
> 
> ```
> Failed tests for configuration config-TU:
> 
>   Check failed with difference files
>   - ../build/test-config-TU/github-0332.xetex.diff
> ```

I'm not sure which branch you are on: I've just tested here and all is well on both `develop` and `master`.


>     * in `./required/tools/build.lua`
> 
> 
> ```
> installfiles = {"*.def", "*.sty", "*.tex", ".tex"}
> ```
> 
> On the l3build side, the change is in the last commit.
> The idea is to only accept dot files when the glob does require it (just like standard globs).

To me, that looks non-standard: `*.tex` means 'all files with the `.tex` extension', and that _includes_ a file called just `.tex`. (Possibly this is because I have a Windows background so me to hidden files are nothing to do with names, whereas extensions are clearly present.)

josephwright:
I guess what I'm still not seeing is what files you are worried about. If you specify `*.tex`, the only issue that could arise is if you had a file called `.tex` that you'd forgotten about as it was being treated as hidden. Something like `.DS_Store` could only match `*.DS_Store`, which looks pretty-much like an opt-in to me.

jlaurens:
Concerning the latex test failure, maybe my TL does not fulfill the requirements, here is the culprit

```
*** ../build/test-config-TU/github-0332.tlg	2021-02-10 12:30:26.000000000 +0100
--- ../build/test-config-TU/github-0332.xetex.log	2021-02-10 12:30:33.000000000 +0100
***************
*** 1,5 ****
--- 1,8 ----
  This is a generated file for the LaTeX2e validation system.
  Don't change this file in any respect.
+ (../se-ascii-print.def
+ File: se-ascii-print.def ....-..-.. v... stringenc: Printable ASCII characters
+ )
  Completed box being shipped out [1]
  \vbox(633.0+0.0)x407.0
  .\hbox(0.0+0.0)x0.0
```

For the other problem, if you specify "\*.tex", not only you have ".tex" but you also have any ".\*.tex".
If you specify "\*\.*" you have absolutely everything that contains a dot, from start to end.
I do not see how to filter out the names with a leading dot.

on unix all dot names are considered special, not only "." and "..".
Moreover, the notion of file extension does not apply to those,
except formally of course.

josephwright:
> Concerning the latex test failure, maybe my TL does not fulfill the requirements, here is the culprit
> 
> ```
> *** ../build/test-config-TU/github-0332.tlg	2021-02-10 12:30:26.000000000 +0100
> --- ../build/test-config-TU/github-0332.xetex.log	2021-02-10 12:30:33.000000000 +0100
> ***************
> *** 1,5 ****
> --- 1,8 ----
>   This is a generated file for the LaTeX2e validation system.
>   Don't change this file in any respect.
> + (../se-ascii-print.def
> + File: se-ascii-print.def ....-..-.. v... stringenc: Printable ASCII characters
> + )

That is _very_ strange: could you send me the 'raw' log file?

> For the other problem, if you specify "*.tex", not only you have ".tex" but you also have any ".*.tex".

Quite true.

> If you specify "*.*" you have absolutely everything that contains a dot, from start to end.

Well yes, `*.*` means 'everything in this directory'.

> on unix all dot names are considered special, not only "." and "..".
> Moreover, the notion of file extension does not apply to those,
> except formally of course.

Both parts true but really I don't see where the real issue is here: one can always list files one-by-one if 'paranoid'. (Even for the 2e kernel we don't, as it gets very painful, but that's the only way to completely exclude any stray files.)

davidcarlisle:
> on unix all dot names are considered special, not only "." and "..".

by shells but not by all unix commands, for example

`find tools -name '*.tex' `

will find all the .tex files in tools including `.tex` (find --help explicitly notes that `*` matches leading dot here)

Changing the behaviour of l3build here is a breaking change for no clear benefit and I don't think we should do it.

FrankMittelbach:
I think the conclusion here is that from our prespective a change in bahvior and interpretation is counter-productive, so from my end this is a clear no. There might be a need to document the current behavior better in the documentation but not to change it.

jlaurens:
It is a little frustrating when you do not know if it really breaks latex2e, you did not ask how many packages would break, and you did not ask what the cost is.
The denial of the fact that dot files are special on unix undermines the cross platform nature of l3build.
Finally I have been criticized for not to be sure that latex2e does not break. It is rather annoying when today latex2e does not always pass the tests and l3build cannot even build the ctan.zip.

davidcarlisle:
as Joseph says it is odd that the test failed for you,  se-ascii-print.def isn't included when I try running that test, do you have a local hyperref or other configuration that is being input?

But I do not see why you should be "frustrated" or "annoyed" you are sending a vast number of PR to a small number of people, for whom the main focus is supporting latex (in volunteer time).  l3build is a tool to that end, not the core product that can take hours of maintenance.  So the only possible outcome is that the PR either breaks latex or doesn't, there are no real advantages for the diffs that are not fixing bugs.  Joseph (mainly) has checked and merged many of your PR but (as you will have seen) testing changes against our current builds is time consuming, so there has to be some clear benefit.  making a breaking change in the filename matching just to make it more like some unix programs and less like some windows ones, and requiring changes to existing build files  just doesn't meet that bar.

jlaurens:
I understand.
In short, correct me if I am wrong, it's impossible to send a PR into anything but master because you refuse to create dev branches and it's impossible to send anything but bug fixes to master.
BTW, I really liked your "some program", extremely funny.

josephwright:
> In short, correct me if I am wrong, it's impossible to send a PR into anything but master because you refuse to create dev branches and it's impossible to send anything but bug fixes to master.

That's not at all how I see it. Over time `l3build` has evolved to add various features and to refactor/revise other ones. The key though has been that they need to contribute in some way to development/testing of LaTeX code, which in rough order of priority is

- `l3build` itself
- The LaTeX kernel
- `expl3`
- Other parts of the `latex3` repo
- Other team-maintained packages
- Other packages, with somewhat higher priority given when they are very widely used

So for example the entire `manifest` set up was written by @wspr as he felt it was useful, even though it was not needed/used by the team directly.

What's really not clear is the wider motivation here. I am trying to follow each PR, but of course the 'fixes a clear bug' ones are much easier to track than the 'changes behaviour without an obvious end goal' ones. We can make a `development` branch, but there's still the question of what the 'plan' is. Moreover, as at any stage `l3build` doesn't have the stability requirements of most LaTeX packages, it can be a bit more flexible in terms of release: what's vital is we know about any changes needed in `build.lua` scripts, not that they don't change.

josephwright:
> The denial of the fact that dot files are special on unix undermines the cross platform nature of l3build.

The reason we are wary on hidden files is that in the past I've messed up including `.tex`. The background here is that I have basically a Windows outlook, as that's my main experience, and in particular here `*.tex` to me says 'match all files with the `.tex` extension', which includes `.tex` itself. (To be clear, that file is not intended to be hidden: it is a file with no 'name' and only an extension'.) I have to confess that prior to this thread it had never occurred to me that `bash` or whatever would not match `.tex` to `*.tex`, as the fact that dot files are regarded as hidden is a feature of `ls` as far as I understand it.

As David has said, there are ways of 'seeing' dot files from the Unix command line. In `l3build` or indeed any script which needs to process file lists, I think we are much closer to `ls -la > temp.txt` then reading `temp.txt` with a simple search, than anything 'command line like'.

What I'm still missing, I guess, is a real case where there is an issue. `.tex` is a good example of a file that _should not_ be regarded as hidden, and it's possible that exactly the same could come up for any other reasonable extension. For something like `.DS_Store` on macOS, you'd need to have `*.DS_Store`, which at least to me feels more like opting in to copying that file. Moreover, if a user is truly being ultra-cautious, then using wildcards is out anyway: `foo.tex` is safer than `*.tex` as it avoids the case where one has copied in some random file for testing or whatever. So there's a balance of convenience here.

Notably, any change might be closer to `bash` or various other Unix shells, but would definitely be further from the behaviour of Command Prompt on Windows. So again it's not clear cut: one cannot be 'the same' as any particular shell and be platform-neutral. Perhaps best, therefore, would be to document the behaviour?

josephwright:
> It is a little frustrating when you do not know if it really breaks latex2e, you did not ask how many packages would break, and you did not ask what the cost is.

Huh? Excluding `.tex` from `*.tex` definitely breaks the `tools` package as it stands as we know it needs that file.

> Finally I have been criticized for not to be sure that latex2e does not break. It is rather annoying when today latex2e does not always pass the tests and l3build cannot even build the ctan.zip.

That can happen, usually due to some external dependence. The recent issues I've been aware of have been due to `hyperref` changes, which will be fixed very soon. Occasionally we have to wait to know that stuff is not broken, and it can be frustrating but it's simply part of life. I'm not sure why you are seeing a different breakage: that is odd as most of the setup for running things is deliberately isolated. I will take a look at the files you sent me when I get  a chance.

FrankMittelbach:
> That can happen, usually due to some external dependence. The recent issues I've been aware of have been due to `hyperref` changes, which will be fixed very soon. Occasionally we have to wait to know that stuff is not broken, and it can be frustrating but it's simply part of life. I'm not sure why you are seeing a different breakage: that is odd as most of the setup for running things is deliberately isolated. I will take a look at the files you sent me when I get a chance.

that is indeed a problem that is unfortunate but unavoidable and occasionally happens (eg during last week twie (and tomorrow most likely again). But I doubt it is worth trying to analyze @josephwright . don't forget that the CTAN upates and local updates are not perfectly synced. Both use possibly different mirrors and their update cycles can differ up to two days (even if you update your local system on a daily basis). I've see those diffs too during the last days too and they went away when travis and my system got back to a synced level. So basically it was not enough to keep the git repo the state of the local TL could produce the or retain such errors.

jlaurens:
When deliberately excluding in l3build ".tex" file from "\*.tex", and use raw GitHub latex2e, check does not complain whereas there is definitely something wrong.
What is the extra procedure to know that latex2e is then broken because of the missing ".tex"?

What would be sufficient to say that latex2e is not broken, knowing that the locally installed files are exactly the ones installed with raw GitHub data?

davidcarlisle:
If this had been missed we could add more tests but I think you will find that `github-0222` will fail (in latex2e/required/tools)


```
This is a generated file for the LaTeX2e validation system.
Don't change this file in any respect.
Author: Bruno Le Floch, Johannes Braams
Format: LaTeX2e<2020-10-01>
Main Class: article
Package: verbatim
(.tex File ignored) (.tex File ignored)
File: "verb-test.tex"  (verbatim)
[1
] (github-0222-input.aux)
```

`.tex` comes from `fileerr.dtx` in the tools bundle so it's natural that the test for it is in `tools` rather than in `base`, although the fact that this particular test picks it up (rather than a specific fileerr test) looks a bit accidental.

josephwright:
I think we've concluded this will not change.

PR #169, opened 2021-02-11
os.type is not a core Lua variable

jlaurens:
It is an addition of luatex (manual p 63)

### DIFF
--- a/l3build.dtx
+++ b/l3build.dtx
@@ -692,10 +692,10 @@
 % \subsection{Selective running of tests}
 %
 % The variables |includetests| and |excludetests| may be used to select which
-% tests are run: these variables take test \emph{names} not full file names.
+% tests are run: these variables take raw test \emph{names} not full file names.
 % The list of tests in |excludetests| overrides any matches in |includetests|,
 % meaning that tests can be disabled selectively. It also makes it possible
-% to disable test on for example a platform basis: the Lua core variable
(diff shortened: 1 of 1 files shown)


josephwright:
Feels like Lua core to me: we are always talking LuaTeX not some arbitrary Lua implementation.

jlaurens:
Problem:`os.type` is undocumented in the dtx.
googling lua + "os.type" gives nothing at all.
googling texlua+"os.type" gives the official manual but not even on the first hint.

Things are more clear if _G, math.pi, for example are denominated core lua variables whereas os.type is denominated core texlua (or luatex) variable.

josephwright:
For me `lua` means `texlua` so implies the LuaTeX manual for info, but I take the point so I've merged.

cereda:
Hello, top of the morning! :wink: 

`os.type` is not a variable, but a _table field_. And even a table is not a variable, but an _object_.

Note that `os` is available in both `lua` and `texlua`, and `type` is a field name which simply happens to be uninitialized (or even deliberately assigned to `nil`) in the former. One can _easily_ mimic the other, either by referencing or unreferencing the corresponding value, as it's just a mere field manipulation.

So saying that `os.type` is not a core `lua` variable is correct (for some definition of _core_ which I am still unsure). But the catch is, it is not a core `texlua` variable either. The true is, it is not even a variable. :tea:

Issue #165, opened 2021-02-08
file-functions: escapepath weirdness


jlaurens:
A cross platform problem
this escape function reads
```
function escapepath(path)
  if os_type == "windows" then
    local path,count = gsub(path,'"','')
    if count % 2 ~= 0 then
      print("Unbalanced quotes in path")
      exit(0)
    else
      if match(path," ") then
        return '"' .. path .. '"'
      end
      return path
    end
  else
    path = gsub(path,"\\ ","[PATH-SPACE]")
    path = gsub(path," ","\\ ")
    return gsub(path,"%[PATH-SPACE%]","\\ ")
  end
end
```
On windows, paths can contain an even number of '"' characters whereas on unix it can contain any number of such delimiter. This seems weird because `escapepath` is used to escape the paths defined in `build.lua` which should be platform agnostic.

josephwright:
The defaults are, but we have to watch for users ;)

jlaurens:
I don't really understand what is exactly behind the scene.
If this is some kind of error recovery, then it should be done on the unix side too, otherwise the same build.lua may break on the unix side.

josephwright:
> I don't really understand what is exactly behind the scene.
> If this is some kind of error recovery, then it should be done on the unix side too, otherwise the same build.lua may break on the unix side.

The point is we can get
```Lua
some-variable = "../An Awkard Dir"
```
in the `build.lua` file, which shows up _after_ we've set the defaults. Now, there are some places I've not got things right with spaces, but this covers a reasonable number of them.

jlaurens:
Ah, I see.
What puzzled me is that your tests are covering something very different like
```
some_variable = "../An\" Awkard \"Dir"
other_variable = 'foo"bar"baz'
```
It does not seem so awkward to deliberately exclude some "exotic" file names, but some kind of note would help.
Maybe raising an error would be appropriate.

josephwright:
Looking back, the point is that this is the same as the rule we using in `expl3`: *nix might allow `"` chars, but Windows really doesn't, so we don't. I think the logic is correct as-is.

PR #173, opened 2021-02-21
Refactor set epoch cmd, clean locals, bug fix

jlaurens:
Simpler code,
useless locals removed,
missing locals added,
bad file exists call

Some locals have been renamed when they hide outer locals.
This is good practice useful in particular when the whole code does not fit in just one screen.

Some code comments to explain the changes.
You may have to check some `show comment` box in the Files changed pane.

### DIFF
--- a/l3build-aux.lua
+++ b/l3build-aux.lua
@@ -60,19 +60,22 @@ function normalise_epoch(epoch)
   end
 end
 
----CLI command to set the epoch, will be run while checking or typesetting
+---Returns the CLI command (ending with `os_concat`) to set the epoch
+---when forcecheckepoch is true, a void string otherwise.
+---Will be run while checking or typesetting
 ---@param epoch string
 ---@return string
 ---@see check, typesetting
 ---@usage private?
 function set_epoch_cmd(epoch)
-  return
+  return forcecheckepoch and (
     os_setenv .. " SOURCE_DATE_EPOCH=" .. epoch
       .. os_concat ..
     os_setenv .. " SOURCE_DATE_EPOCH_TEX_PRIMITIVES=1"
       .. os_concat ..
     os_setenv .. " FORCE_SOURCE_DATE=1"
       .. os_concat
+  ) or ""
 end
 
 ---Returns the script name depending on the calling sequence.
--- a/l3build-check.lua
+++ b/l3build-check.lua
@@ -435,7 +435,7 @@ local function normalize_lua_log(content,luatex)
       end
     end
     -- Look for another form of \discretionary, replacing a "-"
-    pattern = "^%.+\\discretionary replacing *$"
+    local pattern = "^%.+\\discretionary replacing *$"
     if match(line, pattern) then
       return "", line
     else
--- a/l3build-check.lua
+++ b/l3build-check.lua
@@ -806,18 +806,12 @@ function runtest(name, engine, hide, ext, pdfmode, breakout)
   -- Store secondary files for this engine
   for _,filetype in pairs(auxfiles) do
     for _,file in pairs(filelist(testdir, filetype)) do
-      if match(file,"^" .. name .. ".[^.]+$") then
-        local e7n = match(file, "%.[^.]+$")
-        if e7n ~= lvtext and
-           e7n ~= tlgext and
-           e7n ~= lveext and
-           e7n ~= logext then
-           local newname = gsub(file,"(%.[^.]+)$","." .. engine .. "%1")
-           if fileexists(testdir .. "/" .. newname) then
-             rm(testdir,newname)
-           end
-           ren(testdir,file,newname)
+      if match(file,"^" .. name .. "%.[^.]+$") then
+        local newname = gsub(file,"(%.[^.]+)$","." .. engine .. "%1")
+        if fileexists(testdir .. "/" .. newname) then
+          rmfile(testdir,newname)
         end
+        ren(testdir,file,newname)
       end
     end
   end
--- a/l3build-check.lua
+++ b/l3build-check.lua
@@ -698,7 +698,7 @@ function runtest(name, engine, hide, ext, pdfmode, breakout)
   local checkopts = checkopts
   engine = engine or stdengine
   local binary = engine
-  format = gsub(engine,"tex$",checkformat)
+  local format_a = gsub(engine,"tex$",checkformat)
   -- Special binary/format combos
   if specialformats[checkformat] and next(specialformats[checkformat]) then
     local t = specialformats[checkformat]
@@ -706,12 +706,12 @@ function runtest(name, engine, hide, ext, pdfmode, breakout)
     if t_ngn and next(t_ngn) then
       binary    = t_ngn.binary  or binary
       checkopts = t_ngn.options or checkopts
-      format    = t_ngn.format  or format
+      format_a    = t_ngn.format  or format_a
     end
   end
   -- Finalise format string
-  if format ~= "" then
-    format = " --fmt=" .. format
+  if format_a ~= "" then
+    format_a = " --fmt=" .. format_a
   end
   -- Special casing for XeTeX engine
   if match(engine, "xetex") and not pdfmode then
@@ -768,7 +768,7 @@ function runtest(name, engine, hide, ext, pdfmode, breakout)
       -- Ensure lines are of a known length
       os_setenv .. " max_print_line=" .. maxprintline
         .. os_concat ..
-      binary .. format
+      binary .. format_a
         .. " " .. asciiopt .. " " .. checkopts
         .. setup(lvtfile)
         .. (hide and (" > " .. os_null) or "")
--- a/l3build-check.lua
+++ b/l3build-check.lua
@@ -34,7 +34,7 @@ local luatex_version   = status.luatex_version
 
 local len              = string.len
 local char             = string.char
-local frmt             = string.format
+local str_format       = string.format
 local gmatch           = string.gmatch
 local gsub             = string.gsub
 local match            = string.match
@@ -269,7 +269,7 @@ local function normalize_log(content,engine,errlevels)
     -- tidy up to match pdfTeX if an ASCII engine is in use
     if next(asciiengines) then
       for i = 128, 255 do
-        line = gsub(line, utf8_char(i), "^^" .. frmt("%02x", i))
+        line = gsub(line, utf8_char(i), "^^" .. str_format("%02x", i))
       end
     end
     return line, lastline
@@ -345,7 +345,7 @@ local function normalize_lua_log(content,luatex)
         l,
         m .. " (%-?)%d+%.%d+",
         m .. " %1"
-          .. frmt(
+          .. str_format(
             "%.3f",
             match(line, m .. " %-?(%d+%.%d+)") or 0
           )
@@ -714,13 +714,13 @@ function runtest(name, engine, hide, ext, test_type, breakout)
   local binary = engine
   local format = gsub(engine,"tex$",checkformat)
   -- Special binary/format combos
-  if specialformats[checkformat] and next(specialformats[checkformat]) then
-    local t = specialformats[checkformat]
-    local t_ngn = t[engine]
-    if t_ngn then
-      binary    = t_ngn.binary  or binary
-      format    = t_ngn.format  or format
-      checkopts = t_ngn.options or checkopts
+  local special_check = specialformats[checkformat]
+  if special_check and next(special_check) then
+    local engine_info = special_check[engine]
+    if engine_info then
+      binary    = engine_info.binary  or binary
+      format    = engine_info.format  or format
+      checkopts = engine_info.options or checkopts
     end
   end
   -- Finalise format string
@@ -933,10 +933,8 @@ function check(names)
     end
     -- Actually run the tests
     print("Running checks on")
-    local i = 0
-    for _,name in ipairs(names) do
-      i = i + 1
-      print("  " .. name .. " (" ..  i.. "/" .. #names ..")")
+    for i, name in ipairs(names) do
+      print("  " .. name .. " (" ..  i .. "/" .. #names ..")")
       local errlevel = runcheck(name, hide)
       -- Return value must be 1 not errlevel
       if errlevel ~= 0 then
@@ -1015,7 +1013,7 @@ function save(names)
       return 1
     end
     for _,engine in pairs(engines) do
-      local testengine = ((engine == stdengine and "") or ("." .. engine))
+      local testengine = engine == stdengine and "" or ("." .. engine)
       local out_file = name .. testengine .. test_type.reference
       local gen_file = name .. "." .. engine .. test_type.generated
       print("Creating and copying " .. out_file)
--- a/l3build-check.lua
+++ b/l3build-check.lua
@@ -34,7 +34,7 @@ local luatex_version   = status.luatex_version
 
 local len              = string.len
 local char             = string.char
-local format           = string.format
+local frmt             = string.format
 local gmatch           = string.gmatch
 local gsub             = string.gsub
 local match            = string.match
@@ -269,7 +269,7 @@ local function normalize_log(content,engine,errlevels)
     -- tidy up to match pdfTeX if an ASCII engine is in use
     if next(asciiengines) then
       for i = 128, 255 do
-        line = gsub(line, utf8_char(i), "^^" .. format("%02x", i))
+        line = gsub(line, utf8_char(i), "^^" .. frmt("%02x", i))
       end
     end
     return line, lastline
@@ -345,7 +345,7 @@ local function normalize_lua_log(content,luatex)
         l,
         m .. " (%-?)%d+%.%d+",
         m .. " %1"
-          .. format(
+          .. frmt(
             "%.3f",
             match(line, m .. " %-?(%d+%.%d+)") or 0
           )
@@ -712,20 +712,20 @@ function runtest(name, engine, hide, ext, test_type, breakout)
   local checkopts = checkopts
   engine = engine or stdengine
   local binary = engine
-  local format_a = gsub(engine,"tex$",checkformat)
+  local format = gsub(engine,"tex$",checkformat)
   -- Special binary/format combos
   if specialformats[checkformat] and next(specialformats[checkformat]) then
     local t = specialformats[checkformat]
     local t_ngn = t[engine]
-    if t_ngn and next(t_ngn) then
+    if t_ngn then
       binary    = t_ngn.binary  or binary
+      format    = t_ngn.format  or format
       checkopts = t_ngn.options or checkopts
-      format_a    = t_ngn.format  or format_a
     end
   end
   -- Finalise format string
-  if format_a ~= "" then
-    format_a = " --fmt=" .. format_a
+  if format ~= "" then
+    format = " --fmt=" .. format
   end
   -- Special casing for XeTeX engine
   if match(engine, "xetex") and test_type.generated ~= pdfext then
@@ -780,7 +780,7 @@ function runtest(name, engine, hide, ext, test_type, breakout)
       -- Ensure lines are of a known length
       os_setenv .. " max_print_line=" .. maxprintline
         .. os_concat ..
-      binary .. format_a
+      binary .. format
         .. " " .. asciiopt .. " " .. checkopts
         .. setup(lvtfile)
         .. (hide and (" > " .. os_null) or "")
--- a/l3build-check.lua
+++ b/l3build-check.lua
@@ -764,7 +764,7 @@ function runtest(name, engine, hide, ext, pdfmode, breakout)
       -- Allow for local texmf files
       os_setenv .. " TEXMFCNF=." .. os_pathsep
         .. os_concat ..
-      (forcecheckepoch and set_epoch_cmd(epoch) or "") ..
+      set_epoch_cmd(epoch) ..
       -- Ensure lines are of a known length
       os_setenv .. " max_print_line=" .. maxprintline
         .. os_concat ..
--- a/l3build-check.lua
+++ b/l3build-check.lua
@@ -813,7 +813,7 @@ function runtest(name, engine, hide, ext, pdfmode, breakout)
            e7n ~= lveext and
            e7n ~= logext then
            local newname = gsub(file,"(%.[^.]+)$","." .. engine .. "%1")
-           if fileexists(testdir,newname) then
+           if fileexists(testdir .. "/" .. newname) then
              rm(testdir,newname)
            end
            ren(testdir,file,newname)
--- a/l3build-check.lua
+++ b/l3build-check.lua
@@ -1015,7 +1015,7 @@ function save(names)
       return 1
     end
     for _,engine in pairs(engines) do
-      local testengine = ((engine == stdengine and "") or "." .. engine)
+      local testengine = ((engine == stdengine and "") or ("." .. engine))
       local out_file = name .. testengine .. test_type.reference
       local gen_file = name .. "." .. engine .. test_type.generated
       print("Creating and copying " .. out_file)
--- a/l3build-check.lua
+++ b/l3build-check.lua
@@ -134,7 +134,7 @@ local function normalize_log(content,engine,errlevels)
        not match(line, "%.%.%.$") then
       return "", (lastline or "") .. line
     end
-    local line = (lastline or "") .. line
+    line = (lastline or "") .. line
     lastline = ""
     -- Zap ./ at begin of filename
     line = gsub(line, "%(%.%/", "(")
@@ -696,17 +696,17 @@ function runtest(name, engine, hide, ext, pdfmode, breakout)
   cp(lvtfile, fileexists(testfiledir .. "/" .. lvtfile)
     and testfiledir or unpackdir, testdir)
   local checkopts = checkopts
-  local engine = engine or stdengine
+  engine = engine or stdengine
   local binary = engine
-  local format = gsub(engine,"tex$",checkformat)
+  format = gsub(engine,"tex$",checkformat)
   -- Special binary/format combos
   if specialformats[checkformat] and next(specialformats[checkformat]) then
     local t = specialformats[checkformat]
-    if t[engine] and next(t[engine]) then
-      local t = t[engine]
-      binary    = t.binary  or binary
-      checkopts = t.options or checkopts
-      format    = t.format  or format
+    local t_ngn = t[engine]
+    if t_ngn and next(t_ngn) then
+      binary    = t_ngn.binary  or binary
+      checkopts = t_ngn.options or checkopts
+      format    = t_ngn.format  or format
     end
   end
   -- Finalise format string
@@ -807,11 +807,11 @@ function runtest(name, engine, hide, ext, pdfmode, breakout)
   for _,filetype in pairs(auxfiles) do
     for _,file in pairs(filelist(testdir, filetype)) do
       if match(file,"^" .. name .. ".[^.]+$") then
-        local ext = match(file, "%.[^.]+$")
-        if ext ~= lvtext and
-           ext ~= tlgext and
-           ext ~= lveext and
-           ext ~= logext then
+        local e7n = match(file, "%.[^.]+$")
+        if e7n ~= lvtext and
+           e7n ~= tlgext and
+           e7n ~= lveext and
+           e7n ~= logext then
            local newname = gsub(file,"(%.[^.]+)$","." .. engine .. "%1")
            if fileexists(testdir,newname) then
              rm(testdir,newname)
@@ -908,17 +908,12 @@ function check(names)
         end
       end
     end
-    -- https://stackoverflow.com/a/32167188
-    local function shuffle(tbl)
-      local len, random = #tbl, rnd
-      for i = len, 2, -1 do
-          local j = random(1, i)
-          tbl[i], tbl[j] = tbl[j], tbl[i]
-      end
-      return tbl
-    end
     if options["shuffle"] then
-      names = shuffle(names)
+      -- https://stackoverflow.com/a/32167188
+      for i = #names, 2, -1 do
+        local j = rnd(1, i)
+        names[i], names[j] = names[j], names[i]
+      end
     end
     -- Actually run the tests
     print("Running checks on")
--- a/l3build-file-functions.lua
+++ b/l3build-file-functions.lua
@@ -36,7 +36,6 @@ local execute          = os.execute
 local exit             = os.exit
 local getenv           = os.getenv
 local remove           = os.remove
-local os_time          = os.time
 local os_type          = os.type
 
 local luatex_revision  = status.luatex_revision
@@ -183,6 +182,7 @@ function abspath(path)
   error(msg)
 end
 
+-- TODO: Fix the cross platform problem
 function escapepath(path)
   if os_type == "windows" then
     local path,count = gsub(path,'"','')
@@ -214,22 +214,21 @@ end
 -- Copy files 'quietly'
 function cp(glob, source, dest)
   local errorlevel
-  for i,_ in pairs(tree(source, glob)) do
-    local source = source .. "/" .. i
+  for p_rel,p_cwd in pairs(tree(source, glob)) do
     if os_type == "windows" then
-      if attributes(source)["mode"] == "directory" then
+      if attributes(p_cwd)["mode"] == "directory" then
         errorlevel = execute(
-          'xcopy /y /e /i "' .. unix_to_win(source) .. '" "'
-             .. unix_to_win(dest .. '/' .. i) .. '" > nul'
+          'xcopy /y /e /i "' .. unix_to_win(p_cwd) .. '" "'
+             .. unix_to_win(dest .. '/' .. p_rel) .. '" > nul'
         )
       else
         errorlevel = execute(
-          'xcopy /y "' .. unix_to_win(source) .. '" "'
+          'xcopy /y "' .. unix_to_win(p_cwd) .. '" "'
              .. unix_to_win(dest .. '/') .. '" > nul'
         )
       end
     else
-      errorlevel = execute("cp -RLf '" .. source .. "' '" .. dest .. "'")
+      errorlevel = execute("cp -RLf '" .. p_cwd .. "' '" .. dest .. "'")
     end
     if errorlevel ~=0 then
       return errorlevel
@@ -302,10 +301,10 @@ function tree(path, glob)
   end
   local dirs = {["."] = cropdots(path)}
   for pattern, criterion in gmatch(cropdots(glob), "([^/]+)(/?)") do
-    local criterion = criterion == "/" and is_dir or always_true
-    local function fill(path, dir, table)
+    criterion = criterion == "/" and is_dir or always_true
+    local function fill(path_a, dir, table)
       for _, file in ipairs(filelist(dir, pattern)) do
-        local fullpath = path .. "/" .. file
+        local fullpath = path_a .. "/" .. file
         if file ~= "." and file ~= ".." and
           fullpath ~= builddir
         then
@@ -319,13 +318,13 @@ function tree(path, glob)
     local newdirs = {}
     if pattern == "**" then
       while true do
-        local path, dir = next(dirs)
-        if not path then
+        local path_a, dir = next(dirs)
+        if not path_a then
           break
         end
-        dirs[path] = nil
-        newdirs[path] = dir
-        fill(path, dir, dirs)
+        dirs[path_a] = nil
+        newdirs[path_a] = dir
+        fill(path_a, dir, dirs)
       end
     else
       for path, dir in pairs(dirs) do
@@ -357,7 +356,7 @@ function mkdir(dir)
   if os_type == "windows" then
     -- Windows (with the extensions) will automatically make directory trees
     -- but issues a warning if the dir already exists: avoid by including a test
-    local dir = unix_to_win(dir)
+    dir = unix_to_win(dir)
     return execute(
       "if not exist "  .. dir .. "\\nul " .. "mkdir " .. dir
     )
@@ -368,10 +367,10 @@ end
 
 -- Rename
 function ren(dir, source, dest)
-  local dir = dir .. "/"
+  dir = dir .. "/"
   if os_type == "windows" then
-    local source = gsub(source, "^%.+/", "")
-    local dest = gsub(dest, "^%.+/", "")
+    source = gsub(source, "^%.+/", "")
+    dest = gsub(dest, "^%.+/", "")
     return execute("ren " .. unix_to_win(dir) .. source .. " " .. dest)
   else
     return execute("mv " .. dir .. source .. " " .. dir .. dest)
--- a/l3build-file-functions.lua
+++ b/l3build-file-functions.lua
@@ -335,8 +335,8 @@ function tree(src_path, glob)
         fill(p_src, p_cwd, result)
       end
     else
-      for p_scr, p_cwd in pairs(result) do
-        fill(p_scr, p_cwd, new_result)
+      for p_src, p_cwd in pairs(result) do
+        fill(p_src, p_cwd, new_result)
       end
     end
     result = new_result
--- a/l3build-file-functions.lua
+++ b/l3build-file-functions.lua
@@ -312,13 +312,13 @@ function tree(src_path, glob)
     ---@param table table
     local function fill(p_src, p_cwd, table)
       for _, file in ipairs(filelist(p_cwd, glob_part)) do
-        p_src = p_src .. "/" .. file
+        local p_src_file = p_src .. "/" .. file
         if file ~= "." and file ~= ".." and
-          p_src ~= builddir -- TODO: ensure that `builddir` is properly formatted
+        p_src_file ~= builddir -- TODO: ensure that `builddir` is properly formatted
         then
-          p_cwd = p_cwd .. "/" .. file
-          if accept(p_cwd) then
-            table[p_src] = p_cwd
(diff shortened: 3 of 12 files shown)


l3build-aux.lua
```diff
@@ -60,19 +60,22 @@ function normalise_epoch(epoch)
...
+---Returns the CLI command (ending with `os_concat`) to set the epoch
+---when forcecheckepoch is true, a void string otherwise.
+---Will be run while checking or typesetting
 ---@param epoch string
 ---@return string
 ---@see check, typesetting
 ---@usage private?
 function set_epoch_cmd(epoch)
-  return
+  return forcecheckepoch and (
```
jlaurens, line 71: All the calls did include the `forcecheckepoch` test.
Moving this test to the definition of `set_epoch_cmd` makes simpler code

l3build-check.lua
```diff
@@ -134,7 +134,7 @@ local function normalize_log(content,engine,errlevels)
        not match(line, "%.%.%.$") then
       return "", (lastline or "") .. line
     end
-    local line = (lastline or "") .. line
```
jlaurens, line 137: `line` is already local

l3build-check.lua
```diff
@@ -696,22 +696,22 @@ function runtest(name, engine, hide, ext, pdfmode, breakout)
   cp(lvtfile, fileexists(testfiledir .. "/" .. lvtfile)
     and testfiledir or unpackdir, testdir)
   local checkopts = checkopts
-  local engine = engine or stdengine
+  engine = engine or stdengine
   local binary = engine
-  local format = gsub(engine,"tex$",checkformat)
+  local format_a = gsub(engine,"tex$",checkformat)
```
jlaurens, line 701: `format` is already used

l3build-check.lua
```diff
@@ -696,22 +696,22 @@ function runtest(name, engine, hide, ext, pdfmode, breakout)
   cp(lvtfile, fileexists(testfiledir .. "/" .. lvtfile)
     and testfiledir or unpackdir, testdir)
   local checkopts = checkopts
-  local engine = engine or stdengine
```
jlaurens, line 699: engine is already local

l3build-check.lua
```diff
@@ -696,22 +696,22 @@ function runtest(name, engine, hide, ext, pdfmode, breakout)
...
+  engine = engine or stdengine
   local binary = engine
-  local format = gsub(engine,"tex$",checkformat)
+  local format_a = gsub(engine,"tex$",checkformat)
   -- Special binary/format combos
   if specialformats[checkformat] and next(specialformats[checkformat]) then
     local t = specialformats[checkformat]
-    if t[engine] and next(t[engine]) then
-      local t = t[engine]
```
jlaurens, line 706: local t hides outer t

l3build-check.lua
```diff
@@ -764,11 +764,11 @@ function runtest(name, engine, hide, ext, pdfmode, breakout)
       -- Allow for local texmf files
       os_setenv .. " TEXMFCNF=." .. os_pathsep
         .. os_concat ..
-      (forcecheckepoch and set_epoch_cmd(epoch) or "") ..
```
jlaurens, line 767: forcecheckepoch test moved to set_epoc_cmd definition

l3build-check.lua
```diff
@@ -807,13 +807,13 @@ function runtest(name, engine, hide, ext, pdfmode, breakout)
   for _,filetype in pairs(auxfiles) do
     for _,file in pairs(filelist(testdir, filetype)) do
       if match(file,"^" .. name .. ".[^.]+$") then
-        local ext = match(file, "%.[^.]+$")
```
jlaurens, line 810: ext already used, renamed to e7n

l3build-check.lua
```diff
@@ -807,13 +807,13 @@ function runtest(name, engine, hide, ext, pdfmode, breakout)
...
-           ext ~= tlgext and
-           ext ~= lveext and
-           ext ~= logext then
+        local e7n = match(file, "%.[^.]+$")
+        if e7n ~= lvtext and
+           e7n ~= tlgext and
+           e7n ~= lveext and
+           e7n ~= logext then
            local newname = gsub(file,"(%.[^.]+)$","." .. engine .. "%1")
-           if fileexists(testdir,newname) then
```
jlaurens, line 816: only one argument to `fileexists`

l3build-check.lua
```diff
@@ -908,17 +908,12 @@ function check(names)
...
-      for i = len, 2, -1 do
-          local j = random(1, i)
-          tbl[i], tbl[j] = tbl[j], tbl[i]
-      end
-      return tbl
-    end
     if options["shuffle"] then
-      names = shuffle(names)
+      -- https://stackoverflow.com/a/32167188
+      for i = #names, 2, -1 do
```
jlaurens, line 913: shorter and more readable

l3build-check.lua
```diff
@@ -908,17 +908,12 @@ function check(names)
...
+        local j = rnd(1, i)
+        names[i], names[j] = names[j], names[i]
+      end
     end
```
jlaurens, line 917: shorter and more readable

l3build-file-functions.lua
```diff
@@ -36,7 +36,6 @@ local execute          = os.execute
 local exit             = os.exit
 local getenv           = os.getenv
 local remove           = os.remove
-local os_time          = os.time
```
jlaurens, line 39: never used

l3build-file-functions.lua
```diff
@@ -214,22 +214,21 @@ end
 -- Copy files 'quietly'
 function cp(glob, source, dest)
   local errorlevel
-  for i,_ in pairs(tree(source, glob)) do
-    local source = source .. "/" .. i
+  for p_rel,p_cwd in pairs(tree(source, glob)) do
```
jlaurens, line 217: the values in `tree(source, glob)` are already relative to the CWD.
`p_rel` may rename in `p_src`?

l3build-file-functions.lua
```diff
@@ -302,10 +301,10 @@ function tree(path, glob)
   end
   local dirs = {["."] = cropdots(path)}
   for pattern, criterion in gmatch(cropdots(glob), "([^/]+)(/?)") do
-    local criterion = criterion == "/" and is_dir or always_true
```
jlaurens, line 305: criterion is already local

l3build-file-functions.lua
```diff
@@ -302,10 +301,10 @@ function tree(path, glob)
...
-    local function fill(path, dir, table)
```
jlaurens, line 306: path is already used

l3build-file-functions.lua
```diff
@@ -357,7 +356,7 @@ function mkdir(dir)
   if os_type == "windows" then
     -- Windows (with the extensions) will automatically make directory trees
     -- but issues a warning if the dir already exists: avoid by including a test
-    local dir = unix_to_win(dir)
```
jlaurens, line 360: already local

l3build-file-functions.lua
```diff
@@ -368,10 +367,10 @@ end
 
 -- Rename
 function ren(dir, source, dest)
-  local dir = dir .. "/"
+  dir = dir .. "/"
   if os_type == "windows" then
-    local source = gsub(source, "^%.+/", "")
-    local dest = gsub(dest, "^%.+/", "")
+    source = gsub(source, "^%.+/", "")
```
jlaurens, line 372: both are already local

l3build-help.lua
```diff
@@ -39,7 +39,7 @@ end
 function help()
   local function setup_list(list)
     local longest = 0
-    for k,v in pairs(list) do
+    for k,_ in pairs(list) do
```
jlaurens, line 42: unused variable `v` rename

l3build-help.lua
```diff
@@ -70,7 +70,7 @@ function help()
   end
   print("")
   print("Valid options are:")
-  local longest,t = setup_list(option_list)
+  longest,t = setup_list(option_list)
```
jlaurens, line 73: already local at line 63

l3build-install.lua
```diff
@@ -97,7 +97,7 @@ function uninstall()
   if errorlevel ~= 0 then return errorlevel end
   -- Finally, clean up special locations
   for _,location in ipairs(tdslocations) do
-    local path,glob = splitpath(location)
```
jlaurens, line 100: `glob` is unused in the block

l3build-install.lua
```diff
@@ -140,11 +140,11 @@ function install_files(target,full,dry_run)
           end
           local matched = false
           for _,location in ipairs(tdslocations) do
-            local path,glob = splitpath(location)
-            local pattern = glob_to_pattern(glob)
```
jlaurens, line 144: `path` and `glob` are already used

l3build-install.lua
```diff
@@ -162,12 +162,12 @@ function install_files(target,full,dry_run)
     if next(filenames) then
       if not dry_run then
         for _,path in pairs(paths) do
-          local dir = target .. "/" .. path
-          if not cleanpaths[dir] then
-            errorlevel = cleandir(dir)
```
jlaurens, line 167: `dir` is already used → rename

l3build-install.lua
```diff
@@ -192,17 +192,17 @@ function install_files(target,full,dry_run)
       exclude = exclude or { }
       dir = dir or currentdir
       local includelist = { }
-      local excludelist = { }
```
jlaurens, line 195: `excludelist` already used → rename to `exclude list_a`

l3build-manifest-setup.lua
```diff
@@ -221,18 +221,16 @@ end
 --]]
 
 manifest_sort_within_match = manifest_sort_within_match or function(files)
-  local f = files
-  table.sort(f)
-  return f
+  table.sort(files)
+  return files
 end
```
jlaurens, line 226: `files` is already a local variable.

l3build-manifest-setup.lua
```diff
@@ -221,18 +221,16 @@ end
...
 
 manifest_sort_within_group = manifest_sort_within_group or function(files)
-  local f = files
   --[[
       -- no-op by default; make your own definition to customise. E.g.:
-      table.sort(f)
+      table.sort(files)
   --]]
-  return f
+  return files
```
jlaurens, line 233: `files` is already a local variable

l3build-stdmain.lua
```diff
@@ -22,6 +22,8 @@ for those people who are interested.
 
 --]]
 
+local lfs = require("lfs")
```
jlaurens, line 25: Could be
`local lfs = lfs`
but this is more "lua-ish"

l3build-tagging.lua
```diff
@@ -51,7 +51,7 @@ local function update_file_tag(file,tagname,tagdate)
   else
     local path = dirname(file)
     ren(path,filename,filename .. ".bak")
-    local f = assert(open(file,"w"))
```
jlaurens, line 54: `f` is already local at line 42 and unused after line 44

l3build-typesetting.lua
```diff
@@ -38,20 +38,20 @@ local os_type = os.type
 function dvitopdf(name, dir, engine, hide)
   run(
     dir,
-    (forcecheckepoch and set_epoch_cmd(epoch) or "") ..
+    set_epoch_cmd(epoch) ..
```
jlaurens, line 41: `forcecheckepoch` test moved to `set_epoch_cmd` definition

l3build-typesetting.lua
```diff
@@ -38,20 +38,20 @@ local os_type = os.type
...
     "dvips " .. name .. dviext
       .. (hide and (" > " .. os_null) or "")
       .. os_concat ..
-   "ps2pdf " .. ps2pdfopt .. name .. psext
+    "ps2pdf " .. ps2pdfopt .. name .. psext
```
jlaurens, line 45: bad alignment

l3build-typesetting.lua
```diff
@@ -38,20 +38,20 @@ local os_type = os.type
...
       .. (hide and (" > " .. os_null) or "")
   )
 end
 
 -- An auxiliary used to set up the environmental variables
 function runcmd(cmd,dir,vars)
-  local dir = dir or "."
-  local dir = abspath(dir)
-  local vars = vars or {}
```
jlaurens, line 54: both are already local

l3build-typesetting.lua
```diff
@@ -68,7 +68,7 @@ function runcmd(cmd,dir,vars)
   for _,var in pairs(vars) do
     env = env .. os_concat .. os_setenv .. " " .. var .. "=" .. envpaths
   end
-  return run(dir,(forcedocepoch and set_epoch_cmd(epoch) or "") .. env .. os_concat .. cmd)
```
jlaurens, line 71: `forcedocepoch` test moved to `set_epoch_cmd ` definition

l3build-typesetting.lua
```diff
@@ -80,7 +80,7 @@ function biber(name,dir)
 end
 
 function bibtex(name,dir)
-  local dir = dir or "."
```
jlaurens, line 83: `dir` is already local

l3build-typesetting.lua
```diff
@@ -105,7 +105,7 @@ function bibtex(name,dir)
 end
 
 function makeindex(name,dir,inext,outext,logext,style)
-  local dir = dir or "."
```
jlaurens, line 108: `dir` is already local

l3build-typesetting.lua
```diff
@@ -119,15 +119,15 @@ function makeindex(name,dir,inext,outext,logext,style)
 end
 
 function tex(file,dir,cmd)
-  local dir = dir or "."
-  local cmd = cmd or typesetexe .. typesetopts
```
jlaurens, line 123: both are already local

l3build-typesetting.lua
```diff
@@ -119,15 +119,15 @@ function makeindex(name,dir,inext,outext,logext,style)
...
+  dir = dir or "."
+  cmd = cmd or typesetexe .. typesetopts
   return runcmd(cmd .. " \"" .. typesetcmds
     .. "\\input " .. file .. "\"",
     dir,{"TEXINPUTS","LUAINPUTS"})
 end
 
 local function typesetpdf(file,dir)
-  local dir = dir or "."
```
jlaurens, line 130: already local

l3build-typesetting.lua
```diff
@@ -141,7 +141,7 @@ local function typesetpdf(file,dir)
     print(" ! Compilation failed")
     return errorlevel
   end
-  pdfname = name .. pdfext
+  local pdfname = name .. pdfext
   rm(docfiledir,pdfname)
   return cp(pdfname,dir,docfiledir)
```
jlaurens, line 146: The value of `pdfname` is never used out of this block such that `pdfname` is in fact local.

l3build-typesetting.lua
```diff
@@ -226,7 +226,7 @@ function doc(files)
             end
             -- Now know if we should typeset this source
             if typeset then
-              local errorlevel = typesetpdf(srcname,path)
```
jlaurens, line 229: already local

l3build-unpack.lua
```diff
@@ -22,8 +22,6 @@ for those people who are interested.
 
 --]]
 
-local execute          = os.execute
```
jlaurens, line 25: never used

l3build-upload.lua
```diff
@@ -78,6 +78,12 @@ end
 -- if upload is anything else, the user will be prompted whether to upload.
 -- For now, this is undocumented. I think I would prefer to keep it always set to ask for the time being.
 
+local ctan_post -- this is private to the module
```
jlaurens, line 81: `ctan_post` is never used outside of this module.

l3build-upload.lua
```diff
@@ -78,6 +78,12 @@ end
...
+
+-- TODO: next is a public global method,
+-- but following functions are semantically local
+-- despite they are declared globally.
+
```
jlaurens, line 86: Some functions should be declared local, but their definition must me moved before they are used.

l3build-upload.lua
```diff
@@ -100,7 +106,7 @@ function upload(tagnames)
 
   uploadconfig.note =   uploadconfig.note  or file_contents(uploadconfig.note_file)
 
-  local tagnames = tagnames or { }
```
jlaurens, line 103: already local

l3build-upload.lua
```diff
@@ -214,7 +220,7 @@ end
 
 function shell(s)
   local h = assert(popen(s, 'r'))
-  t = assert(h:read('*a'))
+  local t = assert(h:read('*a'))
```
jlaurens, line 223: `t` was global

l3build-aux.lua
```diff
@@ -60,19 +60,22 @@ function normalise_epoch(epoch)
...
     os_setenv .. " SOURCE_DATE_EPOCH=" .. epoch
       .. os_concat ..
     os_setenv .. " SOURCE_DATE_EPOCH_TEX_PRIMITIVES=1"
       .. os_concat ..
     os_setenv .. " FORCE_SOURCE_DATE=1"
       .. os_concat
+  ) or ""
```
jlaurens, line 78: All the calls to `set_epoch_cmd` did include the `forcecheckepoch` test.
Moving this test to the definition of `set_epoch_cmd` makes simpler code

l3build-check.lua
```diff
@@ -134,7 +134,7 @@ local function normalize_log(content,engine,errlevels)
        not match(line, "%.%.%.$") then
       return "", (lastline or "") .. line
     end
-    local line = (lastline or "") .. line
+    line = (lastline or "") .. line
     lastline = ""
```
jlaurens, line 138: `line` is already local as function argument

l3build-check.lua
```diff
@@ -134,7 +134,7 @@ local function normalize_log(content,engine,errlevels)
        not match(line, "%.%.%.$") then
       return "", (lastline or "") .. line
     end
-    local line = (lastline or "") .. line
+    line = (lastline or "") .. line
```
jlaurens, line 137: `line` is already local as function argument

l3build-check.lua
```diff
@@ -696,22 +696,22 @@ function runtest(name, engine, hide, ext, pdfmode, breakout)
   cp(lvtfile, fileexists(testfiledir .. "/" .. lvtfile)
     and testfiledir or unpackdir, testdir)
   local checkopts = checkopts
-  local engine = engine or stdengine
+  engine = engine or stdengine
```
jlaurens, line 699: `engine` is already local as function argument

l3build-check.lua
```diff
@@ -696,22 +696,22 @@ function runtest(name, engine, hide, ext, pdfmode, breakout)
...
   local binary = engine
-  local format = gsub(engine,"tex$",checkformat)
+  local format_a = gsub(engine,"tex$",checkformat)
   -- Special binary/format combos
   if specialformats[checkformat] and next(specialformats[checkformat]) then
     local t = specialformats[checkformat]
-    if t[engine] and next(t[engine]) then
```
jlaurens, line 705: Why `next(t[engine])`

l3build-file-functions.lua
```diff
@@ -368,10 +367,10 @@ end
 
 -- Rename
 function ren(dir, source, dest)
-  local dir = dir .. "/"
+  dir = dir .. "/"
   if os_type == "windows" then
-    local source = gsub(source, "^%.+/", "")
-    local dest = gsub(dest, "^%.+/", "")
```
jlaurens, line 374: both are already local
BTW, there is a problem with the '+' in the pattern.
Either it is required, and we are allowed to refer to the parent directory, or useless.

l3build-install.lua
```diff
@@ -97,7 +97,7 @@ function uninstall()
   if errorlevel ~= 0 then return errorlevel end
   -- Finally, clean up special locations
   for _,location in ipairs(tdslocations) do
-    local path,glob = splitpath(location)
+    local path = splitpath(location)
     errorlevel = zapdir(path)
     if errorlevel ~= 0 then return errorlevel end
```
jlaurens, line 102: `glob` is unused in the block

l3build-install.lua
```diff
@@ -97,7 +97,7 @@ function uninstall()
...
   end
```
jlaurens, line 103: `glob` is unused in the block

l3build-install.lua
```diff
@@ -162,12 +162,12 @@ function install_files(target,full,dry_run)
...
-          if not cleanpaths[dir] then
-            errorlevel = cleandir(dir)
+          local dir_a = target .. "/" .. path
+          if not cleanpaths[dir_a] then
+            errorlevel = cleandir(dir_a)
             if errorlevel ~= 0 then return errorlevel end
           end
-          cleanpaths[dir] = true
+          cleanpaths[dir_a] = true
         end
```
jlaurens, line 171: `dir` is already used → rename to `dir_a`

l3build-install.lua
```diff
@@ -192,17 +192,17 @@ function install_files(target,full,dry_run)
...
           end
         end
       end
       for _,glob in pairs(include) do
         for file,_ in pairs(tree(dir,glob)) do
-          if not excludelist[file] then
+          if not excludelist_a[file] then
             insert(includelist, file)
           end
         end
```
jlaurens, line 208: `excludelist` already used → rename to `exclude list_a`

l3build-typesetting.lua
```diff
@@ -38,20 +38,20 @@ local os_type = os.type
...
 end
 
 -- An auxiliary used to set up the environmental variables
 function runcmd(cmd,dir,vars)
-  local dir = dir or "."
-  local dir = abspath(dir)
-  local vars = vars or {}
+  dir = dir or "."
+  dir = abspath(dir)
+  vars = vars or {}
```
jlaurens, line 54: both are already local

l3build-typesetting.lua
```diff
@@ -68,7 +68,7 @@ function runcmd(cmd,dir,vars)
   for _,var in pairs(vars) do
     env = env .. os_concat .. os_setenv .. " " .. var .. "=" .. envpaths
   end
-  return run(dir,(forcedocepoch and set_epoch_cmd(epoch) or "") .. env .. os_concat .. cmd)
+  return run(dir,set_epoch_cmd(epoch) .. env .. os_concat .. cmd)
```
jlaurens, line 71: `forcedocepoch` test moved to `set_epoch_cmd ` definition

l3build-typesetting.lua
```diff
@@ -80,7 +80,7 @@ function biber(name,dir)
 end
 
 function bibtex(name,dir)
-  local dir = dir or "."
+  dir = dir or "."
```
jlaurens, line 83: `dir` is already local

l3build-typesetting.lua
```diff
@@ -105,7 +105,7 @@ function bibtex(name,dir)
 end
 
 function makeindex(name,dir,inext,outext,logext,style)
-  local dir = dir or "."
+  dir = dir or "."
```
jlaurens, line 108: `dir` is already local

l3build-typesetting.lua
```diff
@@ -119,15 +119,15 @@ function makeindex(name,dir,inext,outext,logext,style)
 end
 
 function tex(file,dir,cmd)
-  local dir = dir or "."
-  local cmd = cmd or typesetexe .. typesetopts
+  dir = dir or "."
+  cmd = cmd or typesetexe .. typesetopts
```
jlaurens, line 123: both are already local

l3build-typesetting.lua
```diff
@@ -119,15 +119,15 @@ function makeindex(name,dir,inext,outext,logext,style)
...
   return runcmd(cmd .. " \"" .. typesetcmds
     .. "\\input " .. file .. "\"",
     dir,{"TEXINPUTS","LUAINPUTS"})
 end
 
 local function typesetpdf(file,dir)
-  local dir = dir or "."
+  dir = dir or "."
```
jlaurens, line 130: `dir` is already local

l3build-typesetting.lua
```diff
@@ -226,7 +226,7 @@ function doc(files)
             end
             -- Now know if we should typeset this source
             if typeset then
-              local errorlevel = typesetpdf(srcname,path)
+              errorlevel = typesetpdf(srcname,path)
```
jlaurens, line 229: already local and used only to return at line 231

l3build-upload.lua
```diff
@@ -100,7 +106,7 @@ function upload(tagnames)
 
   uploadconfig.note =   uploadconfig.note  or file_contents(uploadconfig.note_file)
 
-  local tagnames = tagnames or { }
+  tagnames = tagnames or { }
```
jlaurens, line 109: already local
Better located at line 88

l3build-upload.lua
```diff
@@ -248,11 +254,13 @@ function construct_ctan_post(uploadfile,debug)
   ctan_field("uploader",     uploadconfig.uploader,      255, "Name of uploader",                    true,  false )
   ctan_field("version",      uploadconfig.version,        32, "Package version",                     true,  false )
 
+  --[=[ `qq` is unused because line 258 is commented
```
jlaurens, line 257: line 256/264 instead of line 258
This line # is subject to change

l3build-file-functions.lua
```diff
@@ -302,10 +301,10 @@ function tree(path, glob)
   end
   local dirs = {["."] = cropdots(path)}
   for pattern, criterion in gmatch(cropdots(glob), "([^/]+)(/?)") do
-    local criterion = criterion == "/" and is_dir or always_true
-    local function fill(path, dir, table)
+    criterion = criterion == "/" and is_dir or always_true
```
jlaurens, line 304: criterion is already local

l3build-file-functions.lua
```diff
@@ -302,10 +301,10 @@ function tree(path, glob)
...
+    local function fill(path_a, dir, table)
```
jlaurens, line 305: `criterion` is already local
`path` is already used

l3build-file-functions.lua
```diff
@@ -357,7 +356,7 @@ function mkdir(dir)
   if os_type == "windows" then
     -- Windows (with the extensions) will automatically make directory trees
     -- but issues a warning if the dir already exists: avoid by including a test
-    local dir = unix_to_win(dir)
+    dir = unix_to_win(dir)
```
jlaurens, line 359: already local as function argument

l3build-file-functions.lua
```diff
@@ -368,10 +367,10 @@ end
 
 -- Rename
 function ren(dir, source, dest)
-  local dir = dir .. "/"
+  dir = dir .. "/"
   if os_type == "windows" then
-    local source = gsub(source, "^%.+/", "")
-    local dest = gsub(dest, "^%.+/", "")
+    source = gsub(source, "^%.+/", "")
+    dest = gsub(dest, "^%.+/", "")
```
jlaurens, line 373: both are already local as function arguments
BTW, there is a problem with the '+' in the pattern.
Either it is required, and we are allowed to refer to the parent directory, or useless.

l3build-install.lua
```diff
@@ -140,11 +140,11 @@ function install_files(target,full,dry_run)
           end
           local matched = false
           for _,location in ipairs(tdslocations) do
-            local path,glob = splitpath(location)
-            local pattern = glob_to_pattern(glob)
+            local l_dir,l_glob = splitpath(location)
+            local pattern = glob_to_pattern(l_glob)
```
jlaurens, line 144: `path` and `glob` are already used

l3build-tagging.lua
```diff
@@ -51,7 +51,7 @@ local function update_file_tag(file,tagname,tagdate)
...
+    f = assert(open(file,"w"))
```
jlaurens, line 54: `f` is already local at line 42 and unused after line 44

josephwright:
I'm going to see what @zauguin  and @cereda think about two 'policy' ideas

- Whether one should 'rely' on a `local` already applying to a variable (I've gone with trying to remember to _always_ use `local` even if a variable is already local, but I know there are gaps)
- Whether it's OK to 'hide' outer locals

To me, a lot of the suggested variable names seem less clear - I guess I favoured nicer names over worrying about reuse. Again, that's general so I'll await thoughts from others.

l3build-check.lua
```diff
@@ -435,7 +435,7 @@ local function normalize_lua_log(content,luatex)
       end
     end
     -- Look for another form of \discretionary, replacing a "-"
-    pattern = "^%.+\\discretionary replacing *$"
+    local pattern = "^%.+\\discretionary replacing *$"
```
jlaurens, line 438: `pattern` is not used after line 442.
It is local

zauguin:
>     * Whether one should 'rely' on a `local` already applying to a variable (I've gone with trying to remember to _always_ use `local` even if a variable is already local, but I know there are gaps)

IMO it depends on why a variable is local.

If it was explicitly declared local, I would never redeclare it. That only introduces confusing behavior if you ever get the old one back in scope.

I normally wouldn't explicitly declare function arguments as local, but can make the code clearer for people coming from languages with reference semantics. So I don't have any real opinion on that.

For `for` loop variables I would explicitly declare them as local since changing them (especially the first one) looks as if you are trying to change the (internal) loop control variable of which the first one is a copy. Adding `local` makes it clear that changing the local copy is intentional.

> 
>     * Whether it's OK to 'hide' outer locals
> 
> 
> To me, a lot of the suggested variable names seem less clear - I guess I favoured nicer names over worrying about reuse. Again, that's general so I'll await thoughts from others.

I think it is a good idea to avoid hiding outer locals when reasonable alternatives exist, but I would prefer reused names over unclear ones. But it also depends on how big the area is in which they are used. If the scope which has the reused name has less than ~10 lines, then it's easily visible which variable is meant. If the inner scope doesn't fit on a page, it is very hard to determine this. (IMO it doesn't matter a lot how big the scope of the outer variable is for this.) But in any case, I think that when a variable is renamed because of reuse, we should almost always remove the outer, not the inner variable. After all, this indicates that the name might be too generic and it's less problematic to use generic names for variables with smaller scope.

cereda:
>     * Whether one should 'rely' on a `local` already applying to a variable (I've gone with trying to remember to _always_ use `local` even if a variable is already local, but I know there are gaps)

I'd say we _never_ declare global variables in Lua, we simply assign values to them. That's how the language works. Assigning `nil` to an existing global variable behaves exactly the same as accessing an uninitialized global variable. Their visibility, as the name implies, is global. Similarly, `local` creates scoped variables, bound to the syntactic region they were created, which means their life cycle is limited.

IMHO the real issue at hand (and I am not worried about access benchmarks and whatnot) is _variable shadowing_. Juggling variables back and forth from inner and outer scopes poses as challenging and adds an extra layer of code complexity.

My personal view is to explore function composition as much as possible, leaving assignments and such inside a very limited scope. Also, things that are immutable should be dealt as such (within a read-only table for instance).

josephwright:
> > ```
> > * Whether one should 'rely' on a `local` already applying to a variable (I've gone with trying to remember to _always_ use `local` even if a variable is already local, but I know there are gaps)
> > ```
> 
> I'd say we _never_ declare global variables in Lua, we simply assign values to them. That's how the language works. Assigning `nil` to an existing global variable behaves exactly the same as accessing an uninitialized global variable. Their visibility, as the name implies, is global. Similarly, `local` creates scoped variables, bound to the syntactic region they were created, which means their life cycle is limited.
> 
> IMHO the real issue at hand (and I am not worried about access benchmarks and whatnot) is _variable shadowing_. Juggling variables back and forth from inner and outer scopes poses as challenging and adds an extra layer of code complexity.
> 
> My personal view is to explore function composition as much as possible, leaving assignments and such inside a very limited scope. Also, things that are immutable should be dealt as such (within a read-only table for instance).

Er, for those of us without a CS background, could you say what that means in terms of the PR? ;)

cereda:
> Er, for those of us without a CS background, could you say what that means in terms of the PR? ;)

Sorry. :runner: 

I would surround parts of the code with `do ... end` blocks and/or use `local` whenever possible to ensure nothing survives beyond their original intentions. :wink:

l3build-check.lua
josephwright: From memory, this is checking that table `t` (a) contains an entry for `engine` and (b) that it's a table (so has a valid `next` outcome).

l3build-check.lua
josephwright: `t` is a throw-away variable anyway, so I'm not sure I see the issue with hiding the 'outer' `t`. If we want another name, probably `T` or `tbl` or similar.

l3build-check.lua
josephwright: I really don't like `e7n`, and again I'm not sure `ext` is so bad, but if it needs a new name then `exn` or `extn` or similar - I really don't find `e7n` to be clear. (I guess the 'TeX hat': looks like `e` with an argument `7` and  argument `n` to me)

l3build-check.lua
josephwright: As long as ti works :) I just took the code as-written in the link

l3build-file-functions.lua
josephwright: From memory, this was needed as Windows needs just names here, whereas Unix allows paths.

l3build-install.lua
josephwright: True but it is returned ... to me, a function returning two values needs two places to put them, used or not.

josephwright:
I guess, and this is probably a 'TeX thing', that I actually prefere reusing names as

```lua
local foo
   for ...
    local foo
```

as it avoids adding names to the overall top-level (`_G`), i.e. using space in the hash table

josephwright (commented):
> I think it is a good idea to avoid hiding outer locals when reasonable alternatives exist, but I would prefer reused names over unclear ones. 

I suppose I'm looking at this in TeX terms: if we are thinking of different names, why not just reserve everything in global scope and handle variable use by convention :)

josephwright:
> > ```
> > * Whether one should 'rely' on a `local` already applying to a variable (I've gone with trying to remember to _always_ use `local` even if a variable is already local, but I know there are gaps)
> > ```
> 
> IMO it depends on why a variable is local.
> 
> If it was explicitly declared local, I would never redeclare it. That only introduces confusing behavior if you ever get the old one back in scope.
> 
> I normally wouldn't explicitly declare function arguments as local, but can make the code clearer for people coming from languages with reference semantics. So I don't have any real opinion on that.
> 
> For `for` loop variables I would explicitly declare them as local since changing them (especially the first one) looks as if you are trying to change the (internal) loop control variable of which the first one is a copy. Adding `local` makes it clear that changing the local copy is intentional.

OK, so that's basically a thumbs-up to that aspect of the PR?

l3build-install.lua
jlaurens: The lua 5.3 reference reads
"Before the assignment, the list of values is adjusted to the length of the list of variables. If there are more values than needed, the excess values are thrown away. If there are fewer values than needed, the list is extended with as many nil's as needed."

l3build-check.lua
jlaurens: With the `t_ngn` variable, the problem has disappeared

l3build-check.lua
jlaurens: No problem, e7n means 'e' + 7 letter + 'n'.
I'll go for `extn` instead.

l3build-install.lua
josephwright: Sure, I know it works, I just meant I find it a bit odd.

l3build-file-functions.lua
jlaurens: What puzzles me is "^%.+/" instead of "^%./". If `source` starts with "../" this must be an error, no ?

josephwright: Ah, true

l3build-install.lua
jlaurens: In l3build-ctan.lua line 42, there is the same scheme whereas in l3build-tagging.lua line 52 there is a
`local path = dirname(blablabla)`
I change to the latter in both files.

l3build-check.lua
```diff
@@ -34,7 +34,7 @@ local luatex_version   = status.luatex_version
 
 local len              = string.len
 local char             = string.char
-local format           = string.format
+local frmt             = string.format
```
jlaurens, line 37: The name `format` is reserved for TeX format name.
The function is named after the consonants only (somehow like `rnd` for math.random.)

josephwright:
Any further comments? It looks overall that we are in favour?

jlaurens:
I check the `do ... end` block suggested by cereda and come back as soon as possible

jlaurens:
Finally no `do .. end` block yet. That would make deeper changes more difficult to track down.
- Instead I followed the suggestion to rename the outer variable instead of the inner one when no name alternative comes "naturally". install.lua #165, #173, #190
- `ans` (`answer`) for what is returned install.lua #202, instead of `includelist`
- Added some "(" ")" pair to emphasize the binary operation order (check.lua #1818)
No more change in that PR.

l3build-file-functions.lua
zauguin: Why `p_`? More generally the naming seems off. `p_cwd` looks as if it's the path of the current working directory, but it should be the path to the source file. So it should include some reference to `source` in it's name. IMO `p_src` would be a good fit there since paths are normally assumed to be (absolute or) relative to the CWD, so it's just the "normal" path to the source.

So maybe keeping `p_rel` and replacing `p_cwd` with `p_src`?

jlaurens:
BTW can someone explain why `check.lua` lines 925 (before change) reads 
```
    local i = 0
    for _,name in ipairs(names) do
      i = i + 1
```
instead of just
```
    for i,name in ipairs(names) do
```

josephwright:
> BTW can someone explain why `check.lua` lines 925 (before change) reads
> 
> ```
>     local i = 0
>     for _,name in ipairs(names) do
>       i = i + 1
> ```
> 
> instead of just
> 
> ```
>     for i,name in ipairs(names) do
> ```

The two are different things: `ipairs` is about the table, `i` is about the number of tests. The fact that the table _might_ be numerically indexed is nothing to do with which test we are on.

l3build-file-functions.lua
zauguin: About `path`: I probably don't surprise anyone when I say that I don't like non-descriptive names like `path_a`. `current_path`? `subpath`?

Also: Is there a reason why you didn't change the name in `cropdots` ~10 lines higher? Since that's a one-liner I don't think that it's an issue, but I had the impression that you suggested it pretty consistently everywhere else.

l3build-install.lua
zauguin: I like the new name :+1:

l3build-install.lua
```diff
@@ -200,17 +199,18 @@ function install_files(target,full,dry_run)
           end
         end
       end
+      local ans = { }
       for _,glob in pairs(include) do
         for file,_ in pairs(tree(dir,glob)) do
           if not excludelist[file] then
-            insert(includelist, file)
+            insert(ans, file)
```
zauguin, line 206: 1. YMMV but I would use `res`, `result` or `ret` here. (res and result are already used so they are more consistent, ret shows that it is a generic name for a *ret*urn value. For me `ans` sounds more like a result which is presented to the user.)
  2. The combination of using the more generic name `create_list` and the extremely generic return value name `ans` makes it harder to see on the first glance what this list contains. (Arguably it's still clearer than the old version which claimed the opposite of what it did :laughing:) Maybe `create_filelist`?

l3build-upload.lua
zauguin: I think this should just be removed (including 256/264). If we ever need it again we can try using one of these fancy version control systems to get it back.

zauguin (changes requested):
Mostly looks good, but I have some remaining concerns about naming of things.

l3build-check.lua
```diff
@@ -1020,7 +1015,7 @@ function save(names)
       return 1
     end
     for _,engine in pairs(engines) do
-      local testengine = ((engine == stdengine and "") or "." .. engine)
+      local testengine = ((engine == stdengine and "") or ("." .. engine))
```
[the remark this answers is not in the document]
jlaurens: I personnally prefer the next style which is closer to a ternary operator boolean/value_yes/value_no.
`local testengine = engine == stdengine and "" or ("." .. engine)`
But I did not dare removing extra parenthesis...

jlaurens:
Changes according to almost every zauguin suggestions except:
1) I kept the p_src and p_cwd balance because it is what is output from `tree`. The same path relative to either src or cwd.
2) About the counter problem I mentioned above
```
local i = 0
for _,name in ipairs(names) do
  i = i + 1
  print("  " .. name .. " (" ..  i .. "/" .. #names ..")")
```
and
```
for i, name in ipairs(names) do
  print("  " .. name .. " (" ..  i .. "/" .. #names ..")")
```
are completely equivalent. `i` is no longer used after the print instruction.

l3build-file-functions.lua
```diff
@@ -288,53 +289,59 @@ function filelist(path, glob)
...
-        fill(path, dir, dirs)
+        result[p_src] = nil
+        new_result[p_src] = p_cwd
+        fill(p_src, p_cwd, result)
       end
     else
-      for path, dir in pairs(dirs) do
-        fill(path, dir, newdirs)
+      for p_scr, p_cwd in pairs(result) do
+        fill(p_scr, p_cwd, new_result)
```
zauguin, line 339: `p_scr` looks like a typo for `p_src`.

zauguin (approved):

josephwright:
@jlaurens Anything outstanding or shall I merge?

jlaurens:
I am OK.

Issue #182, opened 2021-03-17
cleandir is dependent on the ordering


stone-zeng:
For example, I need to create and clean the `generic/ctex/zhmap/` dir. However, due to the following code:

https://github.com/latex3/l3build/blob/f0174fee3b19b1faceb3a59b938cef9c144d6a7e/l3build-install.lua#L164-L171

both `generic/ctex/` and `generic/ctex/zhmap/` are created, and the order is not guaranteed. So if `generic/ctex/` is created after `generic/ctex/zhmap/`, the `zhmap` sub-directory will be removed then.

I'm not sure whether the `cleandir()` function should be changed, or maybe we need to normalize the ordering of `paths` in `install_files()`.

josephwright:
I thought we'd looked at this: I'll check back again.

josephwright:
Oh, I remember: I think I didn't get round to going 'all the way' and listing up every path to create in an ordered way. I'll have to think a bit about it.

PR #184, opened 2021-03-19
Fix issue #182

jlaurens:
- `tree` in file functions is ordered in deep first traversal. At a given level, the order is given by `lfs.dir`.
- cleandir is made in reverse order in install
- in clean the rm is also in reverse order (and more efficient)
- some return values are fixed
- more efficient usage of lfs.attributes

Some review is needed before deeper tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ this project uses date-based 'snapshot' version identifiers.
 
 - Normal Lua function calls (issue #127)
 
+### Fixed
+
+- the home texmf directory is now created before use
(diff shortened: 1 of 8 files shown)


josephwright:
I'll need to read through in detail (unless @zauguin beats me to it), but as with the other recent PR, could you add a CHANGELOG entry here?

jlaurens:
Done.
I did not assume that this PR will ever complete, so I separated the CHANGELOG.md updates.
There will certainly be some conflict between the 2 PRs, so another commit when the first one is merged...
and some stash action of yours...

PR #193, opened 2021-05-19
Generate regenerate instructions

### LINKED ISSUE #192: provide option to generate a list of "regenerate" command lines
As suggested by @u-fischer:

 If running a series of checks it would be rather useful if one could get a list of command line instructions to regenerate the failed tests that could then be directly executed.

There is certainly a danger in that one might just regen things without checking well enough why something has changed, but on  the other hand that is something that can happen anyway and having the instructions ready for execution would be rather nice.

zauguin:
Closes #192.

This is a work in progress, but I would love to get some first feedback.

Especially:

  - What should the option be called? (currently there is no option and the commands are always printed)
  - Wording of the text before the commands. Currently: "To regenerate the test files, run". We might want to warn the user that running these commands without thinking pretty much negates the value of testing...

Open tasks:
  - [ ] Add and document an option to turn this on or off.
  - [ ] Testing...
  - [ ] Summary of all configurations. Currently the commands are printed separately at the end of every configuration, it would probably be useful to repeat at the very end the full list.
  - [ ] What to do about modules?
  - [ ] Handling of engine specific tests. Currently it will only print instructions to regenerate non `stdengine` engines if the `stdengine` test passed. This is often the intended behavior, but in some cases this can require two cycles: First `l3build check` only reports the `l3build save` commands to `stdengine`, then after `l3build save` another run of `l3build check` is needed to report engine-specific `l3build save` commands. In general this is hard to avoid, but it could be reduced to cases where the engine/specific differences are newly introduced. Not sure if it's worth it though.
  - [ ] Update the CHANGELOG.


jlaurens:
Could you explain what is the context in use, the actual output and what is the change you propose.

zauguin:
> Could you explain what is the context in use

For context, see the linked issue.

> the actual output and what is the change you propose.

Since it's a WIP, the output is not fixed yet, but currently it show (if the option is passed)

> ```
> To regenerate the test files, run
>     l3build save -c some-config failed-test1 failed-test2
>     l3build save -c some-config -e luatex failed-test3
> ```

etc. Depending on which test failed.

l3build-check.lua
```diff
@@ -574,6 +574,7 @@ function runcheck(name, hide)
   if options["engine"] then
     checkengines = options["engine"]
   end
+  local failedengines = {}
```
jlaurens, line 577: failed_engines

l3build-check.lua
```diff
@@ -582,6 +583,7 @@ function runcheck(name, hide)
     if errorlevel == 0 then
       return errorlevel
     end
+    failedengines[#failedengines + 1] = engine
```
jlaurens, line 586: At the top of the file:
local push = table.insert

push(failed_engines, engine)

l3build-check.lua
```diff
@@ -936,11 +944,13 @@ function check(names)
     end
     -- Actually run the tests
     print("Running checks on")
+    local failurelist = {}
```
jlaurens, line 947: failure_list

l3build-check.lua
```diff
@@ -952,6 +962,9 @@ function check(names)
     end
     if errorlevel ~= 0 then
       checkdiff()
+      if options["show-save-cmds"] then
+        showsavecommands(failurelist)
```
jlaurens, line 966: name consistency
show_save_cmds

l3build-check.lua
```diff
@@ -968,6 +981,34 @@ function checkdiff()
   print("")
 end
 
+-- A short auxiliary to print the list of differences for check
+function showsavecommands(failurelist)
+  local formatcmds = {}
```
jlaurens, line 986: formatcmds → save_cmds

l3build-check.lua
```diff
@@ -968,6 +981,34 @@ function checkdiff()
...
+  local prefix = "    l3build save"
+  if options.config then
+    prefix = prefix .. " -c " .. options.config[1]
+  end
+  for name, engines in next, failurelist do
```
jlaurens, line 991: consistency: pairs(failure_list)

l3build-check.lua
```diff
@@ -968,6 +981,34 @@ function checkdiff()
...
+        if engine == stdengine then
+          formatcmd = prefix
+        else
+          formatcmd = prefix .. " -e " .. engine
+        end
+      end
+      formatcmds[engine] = formatcmd .. " " .. name
+    end
+  end
+  print("\n  To regenerate the test files, run")
```
jlaurens, line 1005: To regenerate the test files, run the following save commands

l3build.dtx
```diff
@@ -379,6 +379,8 @@
 % \item |--rerun| Run tests without unpacking/set up
 % \item |--show-log-on-error| To be used in addition to \texttt{--halt-on-error} and results
 %   in the full \texttt{.log} file of a failed test to be shown on the console
+% \item |--show-save-cmds| When tests fail, print the \texttt{l3build save} commands needed
+%   to regenerate the tests assuming that the failures were false negatives.
```
jlaurens, line 383: When tests fail, print the \texttt{l3build save} commands needed%   to regenerate them after correction. Useful when writing tests.

jlaurens (changes requested):
Formal suggestions for a really better reading experience.

Some testing scenario might be of interest for this feature.
I mean a description of some minimal folder content,
the command to run and the expected output to demonstrate everything.

l3build-check.lua
zauguin: `failedengines` is more consistent with similar names like `checkengines`, `asciiengines`, etc.

l3build-check.lua
zauguin: Why? `push` is not a ommon name for this in Lua, so if at all I would call it `insert`. (Which already exists)

But since the only difference between them is that `table.insert` is slower and IMO less clear, I don't see the point in changing it.

l3build-check.lua
zauguin: The function almost exclusively uses names without underscores.

l3build-check.lua
zauguin: Consistency favors `show-save-cmds`: It is a l3build option. There are no existing l3build options with underscores and multiple existing options with hyphens (`--dry-run`, `--halt-on-error`, and `--show-log-on-error`)

FrankMittelbach:
> Especially:
> 
>     * What should the option be called? (currently it's `--show-save-cmds`)

I guess one gets used to it :-) . don't particularly like it but have no good alternative suggestion.

>     * Wording of the text before the commands. Currently: "To regenerate the test files, run". We might want to warn the user that running these commands without thinking pretty much negates the value of testing...

given that you make the option already that hard to write, somebody using it should know why. I think it is ok, but of course should be properly discussed in the documentation.
> 
> 
> Open tasks:

>     * [ ]  Summary of all configurations. Currently the commands are printed separately at the end of every configuration, it would probably be useful to repeat at the very end the full list.

very useful indeed. Please provide!

>     * [ ]  What to do about modules?

meaning?

 
>     * [ ]  Handling of engine specific tests. Currently it will only print instructions to regenerate non `stdengine` engines if the `stdengine` test passed. This is often the intended behavior, but in some cases this can require two cycles: First `l3build check` only reports the `l3build save` commands to `stdengine`, then after `l3build save` another run of `l3build check` is needed to report engine-specific `l3build save` commands. In general this is hard to avoid, but it could be reduced to cases where the engine/specific differences are newly introduced. Not sure if it's worth it though.

I'm strongly in favor of not generating engine specific tests if the stdengine has failed!. Reason most often they will be bogus, but once installed you will carrythem around unless you make a check across all the tlgs to see if some of them are identical.

> 
>     * [ ]  Update the CHANGELOG.

plus more documentation in the dtx. I think this varants a small section on its own, telling the user when this is useful and what the dangers are (and that there is this 2 levels)

zauguin:
> > ```
> > * [ ]  Summary of all configurations. Currently the commands are printed separately at the end of every configuration, it would probably be useful to repeat at the very end the full list.
> > ```
> 
> very useful indeed. Please provide!

It's certainly useful, I'm just not sure how to implement it yet since they are different processes, so communicating between them is not so easy. We can probably use some helper files though.

> ```
> * [ ]  What to do about modules?
> ```
> meaning?

E.g. in `latex3` and `latex2e`, we can run `l3build check` on the top level and it will got through all subdirectories and run `l3build check` in each of them. These nested invocations are not really aware that they are running as part of one outer invocation, therefore it's complicated to make them include the subdirectory name in the `l3build save` invocation. Additionally we probably need to generate commands like

    cd base
    l3build save ...
    l3build save ...
    cd -

but that's complicated because of platform differences. Does `cd -` work everywhere? Maybe `cd ..` would work better? Or invoking a subshell with parens? Do we have to mark somehow that these are not independent commands but have to be executed in the same shell?

But IMO this has a relatively low priority since during development (where this option is mot useful) you are mostly working inside of a single directory, while "global" `l3build check` are often used more for verification etc. where you normally wouldn't use the option anyway.

It might just fall out of the solution for handling configuration anyway though.

> I'm strongly in favor of not generating engine specific tests if the stdengine has failed!

I agree, the point was mostly for the case when an engine-specific test already exists. E.g. if there already is a luatex specific tlg and both the stdengine and the luatex tests fail, then currently only the instructions for stdengine are provided.

l3build-check.lua
jlaurens: Consistency only applies to similar objects.
`checkengines` and friends are global variables also available from the build.lua.
Your `failedengines` has a completely different semantics, and is local.

l3build-check.lua
jlaurens: Do you mean that
`failedengines[#failedengines + 1] = engine`
is more clear that
`push(failed_engines, engine)`
or
`append(failed_engines, engine)`?

push is a common name for that operation.
Concerning the time loss, it is less than epsilon with respect to the whole process.

l3build-check.lua
jlaurens: The suggestion was targeting the last line.
compare
```
if options["show-save-cmds"] then
  showsavecommands(failurelist)
```
with
```
if options["show-save-cmds"] then
  show_save_cmds(failure_list)
```

FrankMittelbach:
Not sure if that gets to complicated but working with this on the 2e repo which close to a 100 file changes (and the fact that it takes ages to run the whole test suite locally) it might be nice if on the first run it
 - reports resave lines for the stdengines (+ those that only fail for the non-std engines)
 - but also suggest a recheck for just the the files then may still have diffs in the second pass

so that on the second pass you only need to run those that may still be wrong but not the whole set

blefloch:
Rebounding on Frank's comment, there is also the case of tests from other engines that were successful for the stdengine.  So the logic should be, for each test, if stdengine failed list it in the to-be-saved list and in the to-be-rechecked list, while if stdengine was successful, list the other failed engines in the to-be-saved list.

l3build-check.lua
zauguin: Technically it is a local variable here :stuck_out_tongue_winking_eye: More seriously, we use both naming schemes in multiple places for local variables and in this situation, especially given that it's basically the subset of `checkengines` which failed, I think that consistency with `checkengines` is more important.

l3build-check.lua
zauguin: Yes I think that it's more clear since it exactly describes the operation and is a common idiom in Lua. Especially since using `push` or `append`, while common names for similar operations in general, are rather uncommon in Lua code since the operation for tables is often either written like this or using `table.insert`. In general I strongly dislike creating copies of standard functions with different names since this makes them harder to identify and therefore makes people wonder why the standard function hasn't been used.

l3build-check.lua
zauguin: Then take the similar comparison to `showfailedlog` or `showfaileddiff`. I am aware that you don't like the coding style in `l3build` and I agree that it isn't particularly consistent. But since changing the existing functions comes with compatibility and stability concerns, we have to accept it and try to name new functions as consistently as possible.

l3build.dtx
zauguin: I wouldn't want to refer to corrections since the typical usecase is for changes to completely independent code which just happens to influence other tests. Anyway a separate documentation section has been added to expandon the usage.

zauguin:
@FrankMittelbach


> plus more documentation in the dtx. I think this varants a small section on its own, telling the user when this is useful and what the dangers are (and that there is this 2 levels)

I added a small section.

> Not sure if that gets to complicated but working with this on the 2e repo which close to a 100 file changes (and the fact that it takes ages to run the whole test suite locally) it might be nice if on the first run it
> 
>     * reports resave lines for the stdengines (+ those that only fail for the non-std engines)
> 
>     * but also suggest a recheck for just the the files then may still have diffs in the second pass
> 
> 
> so that on the second pass you only need to run those that may still be wrong but not the whole set

Done. While it's certainly very useful, it's less confortable than one might expect since especially when multiple configurations are involved multiple `l3build check` commands are generated which all have to executed separately in order to see the output at the end. (Since checking multiple configurations only really works if all tests are checked)

zauguin:
@FrankMittelbach
> > Especially:
> > ```
> > * What should the option be called? (currently it's `--show-save-cmds`)
> > ```
> 
> I guess one gets used to it :-) . don't particularly like it but have no good alternative suggestion.

Did you come up with a alternative in the meantime? Otherwise I think this should be ready.

blefloch:
On 6/4/21 8:29 PM, Marcel Krüger wrote:
> @FrankMittelbach <https://github.com/FrankMittelbach>
> 
>         Especially:
> 
>         |* What should the option be called? (currently it's `--show-save-cmds`) |
> 
>     I guess one gets used to it :-) . don't particularly like it but have no
>     good alternative suggestion.
> 
> Did you come up with a alternative in the meantime? Otherwise I think this
> should be ready.

Maybe just `--show-saves`?

FrankMittelbach:
>         Especially:
> 
>         |* What should the option be called? (currently it's
>         `--show-save-cmds`) |
> 
>     I guess one gets used to it :-) . don't particularly like it but
>     have no good alternative suggestion.
> 
> Did you come up with a alternative in the meantime? Otherwise I think 
> this should be ready.

not really to be honest, but the way I usually do my workflow (using 
sourcetree diffs to look at the differences after generating whatever 
tlgs are likely to have changed), I will probably use it all the time 
(which is why I don't like they safety "here is complicated option name 
because it might be dangerous" that much), eg would rather like

   l3build save -X

or the like, but I'm good

zauguin:
> Maybe just `--show-saves`?

Ok, I changed it to `--show-saves`.



> which is why I don't like they safety "here is complicated option name because it might be dangerous" that much

The intent was not to make a complicated name because it's dangerous but to use a descriptive name which just happened to be long. I wouldn't want to make this the default since it can be dangerous, but when the user passes any option it's IMHO their responsibility not to do something stupid.



> would rather like
> ```
> l3build save -X
> ```

I added `-X` as a short version of `--show-saves`.

blefloch:
On 6/5/21 2:54 PM, Marcel Krüger wrote:
>     would rather like
> 
>     |l3build save -X |
> 
> I added |-X| as a short version of |--show-saves|.

Is it |l3build check -X| then, or do you have some magic to convert |l3build
save -X| to a checking command that produces the list of saves?

I don't think Frank thought very much about which letter to use, rather just
picked it at a whim to stand for "a short option".

Maybe, with the code you now have, it would be possible to implement a |l3build
save --all| (or whatever the option is in fmtutil to regenerate all formats)
that regenerates all existing tlg that need it?

Bruno

zauguin:
> Is it |l3build check -X| then, or do you have some magic to convert |l3build save -X| to a checking command that produces the list of saves?

It''s good that one of is reading properly :see_no_evil: Right, it's `l3build check -X`.

> I don't think Frank thought very much about which letter to use, rather just picked it at a whim to stand for "a short option".

Probably... Alternative ideas for short options? What about `-S`?

> Maybe, with the code you now have, it would be possible to implement a |l3build save --all| (or whatever the option is in fmtutil to regenerate all formats) that regenerates all existing tlg that need it?

I think that that would be better to implement separately. If we reuse this then we would probably end up doing a full test run and then run the save commands afterwards, while implementing `save --all` can use the results of the first test run directly. (In most cases at least.)

blefloch:
On 6/6/21 12:31 AM, Marcel Krüger wrote:
>     I don't think Frank thought very much about which letter to use, rather just
>     picked it at a whim to stand for "a short option".
> 
> Probably... Alternative ideas for short options? What about |-S|?

That would also be my first thought, but I have no strong opinion.  Let's see
what Frank has to say.

>     Maybe, with the code you now have, it would be possible to implement a
>     |l3build save --all| (or whatever the option is in fmtutil to regenerate all
>     formats) that regenerates all existing tlg that need it?
> 
> I think that that would be better to implement separately. If we reuse this then
> we would probably end up doing a full test run and then run the save commands
> afterwards, while implementing |save --all| can use the results of the first
> test run directly. (In most cases at least.)

Indeed, you are right.

FrankMittelbach:
Am 06.06.21 um 00:43 schrieb Bruno Le Floch:
>  > I don't think Frank thought very much about which letter to use, 
> rather just
>  > picked it at a whim to stand for "a short option".
>  >
>  > Probably... Alternative ideas for short options? What about |-S|?
> 
> That would also be my first thought, but I have no strong opinion. Let's see
> what Frank has to say.

My -X was, as Bruno guessed, just to indicate "something short"

   -S for --show-saves

seems fine

PR #194, opened 2021-06-28
Given that I have forgotten that again for the second time ...

FrankMittelbach:
small reminder to run l3build ctan before uploading :-(

### DIFF
--- a/l3build-upload.lua
+++ b/l3build-upload.lua
@@ -116,11 +116,24 @@ function upload(tagnames)
   end
 
   -- avoid lower level error from post command if zip file missing
-  local zip=open(trim_space(tostring(uploadfile)),"r")
-  if zip~=nil then
-    close(zip)
-  else
-    error("Missing zip file '" .. tostring(uploadfile) .. "'")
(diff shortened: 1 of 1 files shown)


PhelypeOleinik (approved):

zauguin:
Maybe we could extend this along the lines of https://github.com/latex3/l3build/compare/warn-about-old-ctan: Instead of always warning, make it dependant on the age of the upload file. E.g. currently it does

  - Not warn when the file is less than a minute old
  - State the age of the file if it's older than a minute but less than an hour old.
  - Print a big warning if it's older than an hour but younger than a day.
  -  Ask for explicit user approval if the file is older than a day.

This makes the probably most common case (the user wasn't aware of having to run `l3build ctan`) harder to ignore, without training people to always ignore the warning.

FrankMittelbach:
> This makes the probably most common case (the user wasn't aware of 
> having to run |l3build ctan|) harder to ignore, without training people 
> to always ignore the warning.

nice idea. I don't really think it is necessary (I would have been 
content with the thing reminding me every time :-) but since you already 
implemented it, I'm good with it, it certainly is an improvement.

PR #198, opened 2021-08-05
Don't rely on external program for zip file generation

### LINKED ISSUE #197: File permissions with l3build ctan
From https://tex.stackexchange.com/q/609646/105447

CTAN states some file permission requirements for uploads which, as far as I can tell, do not seem to be catered for or even controllable from `l3build ctan`.

These requirements are given in https://ctan.org/file/help/ctan/CTAN-upload-addendum#filepermissions, and are:

> a.  Only files that are truly executable (like Shell, Perl, Python, Ruby, and other scripts) should be marked as such. An “executable” README or .tex file does not make any sense.
>  b.  Files submitted to CTAN are obviously meant for *publication*. This implies that they should be “world readable”. (In fact, exaggerated “privacy” of file permissions can lead to serious complications in the installation process.)

However, on an Linux system at least, whatever file permissions are in use are currently honored by `zip` and hence do make into the `.zip` file generated by `l3build ctan`.

So I kindly request that some means to comply with those CTAN upload requirements be provided for `l3build ctan`. Considering the case, probably it should "just do what CTAN expects" (as it already does with end-of-line, for example). Alternatively, some way to control file permissions of the included files, would also be very welcome.

zauguin:
One approach to fix #197: If `zip` doesn't allow us to set the permissions correctly, simply don't use it and generate ZIP files on our own. 

Needs testing on Windows, especially regarding normalization of EOL markers (that should be handled by opening text files as text files, but things tend to be more complicated in that area).

### DIFF
--- a/l3build-ctan.lua
+++ b/l3build-ctan.lua
@@ -25,8 +25,11 @@ for those people who are interested.
 local pairs = pairs
 local print = print
 
+local attributes = lfs.attributes
 local lower = string.lower
 local match = string.match
+
+local newzip = require"l3build-zip"
 
 -- Copy files to the main CTAN release directory
 function copyctan()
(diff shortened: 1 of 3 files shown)


josephwright:
Doesn't this mean we need to remove `zipopts` and `zipexe`, plus update the docs?

josephwright:
Nice plan, BTW.

Issue #210, opened 2021-11-22
Support files not visible to tex files in unpackdir


leingang:
Copied from [TSX](https://tex.stackexchange.com/q/623511/1402):

I have a module that I want to manage with l3build.  Unpacking a docstrip file generates a tex file.  `l3build doc` typesets that generated tex file.  But I can't figure out how to configure `build.lua` to enable that process to see any images that should be included.

My MWE is four files (all in a [Github repo][1] for ease)

**file tree**

    ├── build.lua
    ├── foo.dtx
    ├── foo.ins
    └── support
        └── duck.png

**`foo.dtx`**

    %<*foo>
    \documentclass{article}
    \usepackage{graphicx}
    \usepackage{lipsum}
    \begin{document}
    \lipsum[1]
    \includegraphics[
        width=\textwidth,
    %<bar>angle=180,
    ]{duck}
    \end{document}
    %</foo>

**`foo.ins`**

    \input docstrip.tex
    \askforoverwritefalse
    \generate{\file{bar.tex}{\from{\jobname.dtx}{foo,bar}}}
    \endbatchfile

So `bar.tex` has one extra line, which should rotate the image 180 degrees.

**`build.lua`**

    module = "foo"
    
    typesetsuppfiles = {"duck.png"}
    typesetfiles = {"foo.dtx","bar.tex"}

**duck.png**

[![duck][2]][2]

Command:

    l3build doc foo

* `duck.png` is copied into `typesetdir`
* `foo.dtx` is typeset in `typesetdir` (I think)
* result: success.  `foo.pdf` is moved into `maindir`.

Command:

    l3build doc bar

* `bar.tex` is unpacked into `unpackdir`
* `tex` is run on `bar.tex` in `unpackdir`
* result: failure because `duck.png` is not found.

  [1]: https://github.com/leingang/test-l3build-doc-images/tree/master
  [2]: https://i.stack.imgur.com/BkFhn.png

josephwright:
I guess either @zauguin or I will handle it

josephwright:
Fixed by ff20ff23b3da22bf32530623a5dddbdb08ef6360

PR #211, opened 2021-11-22
Always typeset in typesetdir

zauguin:
Suggested fix for #210: Instead of typesetting unpacked files directly in unpackdir, copy them to typesetdir first.

Additionally reorganizes the setup steps to ensure that we still prefer direct files over unpacked files if both match.

### DIFF
--- a/.github/tl_packages
+++ b/.github/tl_packages
@@ -33,6 +33,7 @@ gettitlestring
 graphics
 hologo
 hycolor
+hypdoc
 iftex
 intcalc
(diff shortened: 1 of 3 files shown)


josephwright:
Looks good to me - other than the test failures!

zauguin:
> other than the test failures

We get the same failures on `main` and they all seem to be about 0pt kerns, so they seem unrelated. Maybe they are caused by the 2e update?

PR #213, opened 2021-11-25
double backslashes when writing to curlopts file

davidcarlisle:
double up `\ `  when writing the curlopts file (leaving `\n` for newline etc) so you do not have to double `\ ` in files or on the terminal (in fields entered as Lua strings still need to use `\\ ` or use the `[[ ...]]`  unescaped string literals.)

This is a backwards incompatible change in that existing sumary or announcement strings will get doubled `\\ ` but announcememts at least are hopefully one-off strings not so this should not be too bad and at worst you get a slightly garbled message; the upload should still work.


josephwright:
CHANGELOG and doc entry?

FrankMittelbach:
I would say yes to both, other than that I think it is a good move

josephwright (approved):

Issue #205, opened 2021-09-02
setting checkinit_hook breaks tests


u-fischer:
the hyperref build.lua contains a definition for `checkinit_hook ()` and this now breaks the test runs:  the files are unpackage but then the tests don't start.

Removing the function solved the problem. 

I could also reproduce it in tagpdf. A simple

~~~~
function checkinit_hook ()
 print("XXXXXXXXXXXXXX")
end
~~~~

stopped the tests there too.

josephwright:
Did you try with a return value? I think you have to have `return 0` here.

u-fischer:
I did now, and this seems to work. But something must have changed as the tests passed in june at the last hyperref update.

I would suggest to document that a return value 0 (or whatever) is required. Currently the documentation only says that the hook is there.

zauguin:
This change was the fix for #189, there the issue was that errors reported by `checkinit_hook` through the return value were ignored.

Do we want to fix this by updating the documentation or by treating the hook as successful if it doesn't return anything?

josephwright:
The docs do cover this: I think we can close.

PR #219, opened 2021-12-10
stdmain: allow a target returning `nil`

muzimuzhi:
When [digging `pgf`'s CI](https://github.com/pgf-tikz/pgf/pull/1094#discussion_r766660074), I find currently, if a (user-declared) target returns `nil`, the `main()` in `stdmain` will call `exit(1)`. Since `nil` is the default return value of a Lua function, accepting it would be better.

This PR makes the `main()` call `exit(0)` when a target returns `nil`.

PS: Should I add a testfile?

```bash
$ cat build.lua
```
```lua
module = "l3buid-test"

target_list.myfun =
  {
    func = function() print("This is myfun2().") end
  }
```
```bash
# expected: "Exit code is 0"
# current: "Exit code is not 0"
$ l3build myfun && echo "Exit code is 0" || echo "Exit code is not 0"
This is myfun2().
Exit code is not 0
```

### DIFF
--- a/l3build-stdmain.lua
+++ b/l3build-stdmain.lua
@@ -179,7 +179,7 @@ function main(target,names)
     errorlevel = target_list[target].func(names)
   end
   -- All done, finish up
-  if errorlevel ~= 0 then
+  if errorlevel or errorlevel ~= 0 then
     exit(1)
   else
     exit(0)


l3build-stdmain.lua
```diff
@@ -179,7 +179,7 @@ function main(target,names)
     errorlevel = target_list[target].func(names)
   end
   -- All done, finish up
-  if errorlevel ~= 0 then
+  if errorlevel or errorlevel ~= 0 then
```
zauguin, line 182: `s/or/and/`

zauguin (changes requested):
Beside that the current code is wrong, I also disagree on a more fundamental level with this change.

By design we use the return value to indicate success or failure. While we use numeric error codes, the other conventions for doing this would be to return boolean or (especially when taking Lua 5.4 conventions into account) to return `fail` (aka. `nil`) for failure conditions.

While this change would make returning nothing work, it would break code using  these other patterns. Also making `return fail` (or even `return false`) end up indicating success is IMO worse than to indicate failure when the return statement has been forgotten (especially since error paths are usually less tested).

l3build-stdmain.lua
muzimuzhi: Perhaps I should close this PR, and suggest `l3build` to add words to the doc for `target_list`, to mention that the `func` should return 0 when successes.
https://github.com/latex3/l3build/blob/177e77d514df4fc139a47e889344833ebda0645e/l3build.dtx#L1931-L1934

u-fischer: Is that the same as this https://github.com/latex3/l3build/issues/205?

muzimuzhi: https://github.com/latex3/l3build/issues/205#issuecomment-977680131 makes me re-check the latest manual, and this time I find the line 
https://github.com/latex3/l3build/blob/177e77d514df4fc139a47e889344833ebda0645e/l3build.dtx#L1946

So it's totally my careless. I'm closing this PR.

PR #220, opened 2021-12-17
use form-string not form to guard against leading < or @

davidcarlisle:
the curl manual says

```
--form-string <name=string>

(HTTP SMTP IMAP) Similar to -F, --form except that the value string for the named parameter is used literally.
Leading '@' and '<' characters, and the ';type=' string in the value have no special meaning.
Use this in preference to -F, --form if there's any possibility that the string value may accidentally trigger
the '@' or '<' features of -F, --form. 
```

and eg the pgf description starts with `<p>` and the leading `<` breaks everything.

This uses `form-string` instead of `form` for all fields except the zip file upload that needs the leading `@` to be interpreted.

### DIFF
--- a/l3build-upload.lua
+++ b/l3build-upload.lua
@@ -324,7 +324,7 @@ function ctan_single_field(fname,fvalue,max,desc,mandatory)
       vs = vs:gsub('`','\\`')
       vs = vs:gsub('\n','\\n')
 -- for strings on commandline version      ctan_post=ctan_post .. ' --form "' .. fname .. "=" .. vs .. '"'
-      ctan_post=ctan_post .. '\nform="' .. fname .. '=' .. vs .. '"'
+      ctan_post=ctan_post .. '\nform-string="' .. fname .. '=' .. vs .. '"'
     end
   else
     error("The value of the field '" .. fname .."' must be a scalar not a table")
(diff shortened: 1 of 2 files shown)


josephwright:
CHANGELOG ...

hmenke (approved):

davidcarlisle:
I decided to check the error status of (all) shell calls here so converting to draft while preparing a second commit for this PR

davidcarlisle:
Draft status lifted:-)

l3build-upload.lua
```diff
@@ -245,8 +245,12 @@ end
...
   local h = assert(popen(s, 'r'))
   local t = assert(h:read('*a'))
-  h:close()
-  return t
+  local success = h:close()
+  if (success) then
+   return t
+  else
+   error("\nError from shell command:\n" .. s .. "\n" .. t .. "\n")
+  end
```
hmenke, line 253:
```suggestion
  h:close() or error("\nError from shell command:\n" .. s .. "\n" .. t .. "\n")
  return t
```

davidcarlisle: @hmenke  more idiomatic Lua I guess, happy to change to that

davidcarlisle: @hmenke 

```
error loading module ./l3build-upload.lua from file ./l3build-upload.lua:
        ./l3build-upload.lua:248: unexpected symbol near 'or'

```

it doesn't seem to like inlining h:close (which has three return values) here?

hmenke: Nevermind then, it was just a thought.

Issue #228, opened 2022-04-15
Install files into directories with spaces in their names


leingang:
I am using `l3build` to automate document generation (not package testing or installation). I would like to install the documents in a documents tree rather than in a `texmf` tree.

The workaround I'm using is to set `options['texmfhome']` to the directory I want to install the files in, and `tdslocations` to install any PDFs generated at the top of this directory, like so:

    options['texmfhome'] = '/path/to/install'
    sourcefiles = {}
    tdslocations = {'*.*.pdf'}

This works pretty well, _unless_ the path in the first line leads to a directory with spaces in it.  For instance, if `options['texmfhome']` is set to `/Users/mathew/Public/foo bar', the error I get is:

    cp: /Users/matthew/Public/foo\ bar/.: No such file or directory

I know that spaces in paths have been an issue in the past (e.g., issue #124), and I realize that a “real” `texmf` tree probably shouldn't have a space in its name, but I hope this could be addressed.

josephwright:
Unless @zauguin  or @PhelypeOleinik  fancy it, I guess I need to fix this (or rope @cereda into it!)

leingang:
Thank for you looking into this so quickly! I just installed the new version and it works for me!

PR #229, opened 2022-04-19
Fix issue when installing into directory with spaces

zauguin:
Fix for #228.

There are three independent issues here:

  1. `mkdir` doesn't escape it's argument. Fixed by adding `escapepath`.
  2. `escapepath` is broken on non-Windows for pathes with spaces since the pattern doesn't take care of `-` being a special character. Fixed by adding a `%`.
  3. `cp` always gets an escaped path passed as destination but quotes the destination argument again. For now this PR removes the quoting, but in the longer term I think that this should be fixed the other way around.

Generally while this change allows for changes in tdshome, the handling of spaces in filenames needs much more fundamental work.

### DIFF
--- a/l3build-file-functions.lua
+++ b/l3build-file-functions.lua
@@ -198,7 +198,7 @@ function escapepath(path)
   else
     path = gsub(path,"\\ ","[PATH-SPACE]")
     path = gsub(path," ","\\ ")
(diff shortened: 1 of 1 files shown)


josephwright:
Great, thanks: I agree we still need more work overall, but this should fix the current issue.

PR #235, opened 2022-06-10
fix!: make sure that typesetsuppfiles are copied into the localdir

teatimeguest:
The [documentation][doc] says that `typesetsuppfiles` are copied into the `localdir`,
but in fact they are copied into the `typesetdir`. This PR fixes this problem.

[doc]: https://github.com/latex3/l3build/blob/87f00fcca1f8db39fefa1f482fb2f47f308db377/l3build.dtx#L513


josephwright:
I think that will work but I wonder if we should instead change the docs: @zauguin thoughts?

josephwright:
Fixed by altering docs

PR #241, opened 2022-09-14
Better handle spaces in file and path names

zauguin:


josephwright:
Seems to miss `l3build-unpack`, and then life gets ... interesting

zauguin:
@josephwright Should be fixed now. Previously I only looked for commands called with `os.execute` and therefore missed the two files using `io.popen`.

josephwright:
Now `unpack`, `install`, `check` all work, but `doc` (and thus `ctan`) fail.

josephwright:
@zauguin  Do you think we can get this working or should we close the PR?

josephwright:
@zauguin I'm minded to close here ATM as we are not sure about this - thoguhts?

PR #259, opened 2022-12-11
typo

yegor256:
There is a typo in documention. `return 0` leads to this runtime error:

```
sh: 0: command not found
```


josephwright:
Looking at it, the bit that's wrong is the other line: it should be `return runcmd("biber " .. name,".")` I think.

PR #262, opened 2023-01-11
Use runcmd in dvitopdf to allow for local texmf

teatimeguest:
The current implementation of `dvitopdf` does not set `TEXMFCNF`, which prevents `dvips` from finding local texmf fonts. The PR fixes this problem by invoking `dvips` via `runcmd` function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ this project uses date-based 'snapshot' version identifiers.
 
 ## [Unreleased]
 
+### Fixed
+- Ensure `texmf.cnf` work correctly for `dvips`
+
 ## [2022-11-10]
 
 ### Changed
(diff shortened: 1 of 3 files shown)


josephwright:
I guess this needs applying generally in the `typeset` file, plus we should re-order so `runcmd()` comes first.

teatimeguest:
I have made those changes.

josephwright:
Cool. Could you include something in the ChangeLog?

teatimeguest:
Will this be OK?

josephwright:
> Will this be OK?

Almost ;) Please put the ChangeLog entry under `Unreleased`: the tag gets done when we actually send to CTAN.

teatimeguest:
Fixed.

PR #267, opened 2023-01-19
Fix #261, add zsh completion

### LINKED ISSUE #261: [feature] Can l3build provide shell completions?
Just like

```sh
❯ l3build --<TAB>
option
--date                Sets the date to insert into sources
--debug               Runs target in debug mode
--dirty               Skip cleaning up the test area
--dry-run             Dry run for install or upload
--email               Email address of CTAN uploader
--epoch               Sets the epoch for tests and typesetting
--first               Name of first test to run
...
```

TIA!

Freed-Wu:
Users must `autoload -Uz compinit && compinit` in `~/.zshrc` to enable
completion. Then

1. Users can add `eval "$(l3build completion zsh)"` to `~/.zshrc`
2. Users can add the following file (will be added to
<https://github.com/zsh-users/zsh-completions>) to `/usr/share/zsh/site-functions/_l3build`:

```zsh

(( $+functions[__l3build] )) || eval "$(l3build complete zsh)" && __l3build
```

Then `grep l3build ~/.zcompdump` to confirm:

```
'l3build' '_l3build'
```

If not, `rm ~/.zcompdump`, then `compinit` again.

Note: `/usr/share` is only for linux, for other OSs, try

- macOS: `/usr/local/share`
- Android (Termux): `/data/data/com.termux/files/usr/share`
- Windows (msys2/mingw64): `/mingw64/share`

### DIFF
--- a/l3build-arguments.lua
+++ b/l3build-arguments.lua
@@ -40,7 +40,8 @@ option_list =
       {
         desc  = "Sets the config(s) used for running tests",
         short = "c",
-        type  = "table"
+        type  = "table",
+        complete = "lua_file"
       },
     date =
       {
@@ -71,7 +72,8 @@ option_list =
       {
         desc  = "Sets the engine(s) to use for running test",
         short = "e",
-        type  = "table"
+        type  = "table",
+        complete = "engine"
       },
     epoch =
       {
@@ -82,7 +84,8 @@ option_list =
       {
         desc  = "Take the upload announcement from the given file",
         short = "F",
-        type  = "string"
+        type  = "string",
+        complete = "file"
       },
     first =
       {
@@ -110,7 +113,8 @@ option_list =
       {
         desc  = "Print this message and exit",
         short = "h",
-        type  = "boolean"
+        type  = "boolean",
+        stop_completions = {"option", "target", "name"}
       },
     last =
       {
@@ -158,7 +162,8 @@ option_list =
     version =
       {
         desc = "Print version information and exit",
-        type = "boolean"
+        type = "boolean",
+        stop_completions = {"option", "target", "name"}
       }
   }
 
--- a/l3build-complete.lua
+++ b/l3build-complete.lua
@@ -1,0 +1,148 @@
+--[[
+
+File l3build-complete.lua Copyright (C) 2018,2020,2021 The LaTeX Project
+
+It may be distributed and/or modified under the conditions of the
+LaTeX Project Public License (LPPL), either version 1.3c of this
+license or (at your option) any later version.  The latest version
+of this license is in the file
+
+   http://www.latex-project.org/lppl.txt
+
+This file is part of the "l3build bundle" (The Work in LPPL)
+and all files in that bundle must be distributed together.
+
+-----------------------------------------------------------------------
+
+The development version of the bundle can be found at
+
+   https://github.com/latex3/l3build
+
+for those people who are interested.
+
+--]]
+
+local insert = table.insert
+local match  = string.match
+local gsub   = string.gsub
+local sort   = table.sort
+local concat = table.concat
+
+function complete(name)
+  if name == "zsh" then
+    complete_zsh()
+  end
+end
+
+function complete_zsh()
+  local zsh_template = [=[
+#compdef {{ scriptname }}
+
+__l3build() {
+  local targets=(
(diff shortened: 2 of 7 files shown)


Freed-Wu:
I only realize a zsh completion. For other shells, it is similar.

l3build-arguments.lua
```diff
@@ -40,12 +40,13 @@ option_list =
...
         short = "c",
-        type  = "table"
+        type  = "table",
+        complete = "lua_file"
       },
     date =
       {
         desc  = "Sets the date to insert into sources",
-        type  = "string"
+        type  = "string",
```
josephwright, line 49: Why the stray comma here?

l3build-arguments.lua
```diff
@@ -65,13 +66,14 @@ option_list =
     email =
       {
         desc = "Email address of CTAN uploader",
-        type = "string"
+        type = "string",
```
josephwright, line 69: Again, an odd stray comma

josephwright (changes requested):
There doesn't seem to be any documentation here: we need something in `l3build.dtx`, probably in the `man` source and also in the ChangeLog.

l3build-arguments.lua
Freed-Wu: It's because I am too tired to check the format at that time. Fixed.

l3build-arguments.lua
Freed-Wu: Fixed.

l3build-complete.lua
```diff
@@ -0,0 +1,133 @@
+--[[
+
+File l3build-help.lua Copyright (C) 2018,2020,2021 The LaTeX Project
```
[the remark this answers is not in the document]
Freed-Wu: It looks like I am truly tired :smile: Fixed.

Freed-Wu:
Fixed.

josephwright:
> BTW, why not `chmod +x l3build.lua`? Because this file has a shebang.

You only need that if you are installing locally on a *Nix system, but then you can do `texlua $(kpsewhich l3build.lua)` anyway ... otherwise, we leave things to Karl for TL.

Freed-Wu:
Is there any other problem about this PR? [Similar issue](https://github.com/TeX-Live/texdoc/issues/89) in [texdoc](https://github.com/TeX-Live/texdoc) have been finished.

josephwright:
Personally, I am not keen - this is a platform-specific feature whilst `l3build` aims to be platform-neutral. However, it's not just my call - thoughts from others?

davidcarlisle:
I think I agree with @josephwright  here.  This seems to be a separate feature that could be installed separately to l3build. If we add zsh where would it end, other *ix shells notably bash have similar features and probably windows powershell and other commandlines . We could end up with a non trivial amount of code and documentation for a system we can't easily test for shells we don't use.

Freed-Wu:
So the reasons to disagree this PR are:

- it will increase code and document

Anyone have other better solution to support completion without increasing many code? Or just give up this feature?

- if this module has bug, no one can maintain this module. because `we can't easily test for shells we don't use.`

IMO, the test and maintenance should be the task of who use this feature, if a shell doesn't have any user to use, the feature will not be added because no one do this work. For those shells many users use, the completion for those shell will have many users to test and maintain.

The completion of `texdoc` have been added by many linux distributions. According to the response of many linux distributions' packager, adding a shell completion is welcome.

josephwright:
@Freed-Wu  I looked the PR when it came in, but being from a background where auto-completing options feels odd, I didn't want to merge. At the time, I was hoping other team members would look and comment. It then slipped off the radar, which is why when you reminded me I put an explicit comment saying what I felt but also seeing if that's a wider view. So it's not been rejected: the question is what the consensus is.

Freed-Wu:
No offence. I just am surprised there is an issue which pass through about one year after some reviews in Jan 20 and want to know its status :smile:

josephwright:
> No offence. I just am surprised there is an issue which pass through about one year after some reviews in Jan 20 and want to know its status 😄

Really sorry this has taken so long to get back on. As I said, it got a bit lost in the overall scheme of things. 

We had a chat about this today at the team meeting, and have decided on balance to close, broadly for the reasons @davidcarlisle  outlined.

PR #269, opened 2023-02-19
Unify `testdir` when config is passed

muzimuzhi:
Using `l3build` repo as an example, currently, 
- `l3build check -c config-pdf` creates `./build/config-pdf` as `testdir`, while
- `l3build check -c config-pdf.lua` creates `./build/config-pdf.lua` as `testdir`, noted the trailing `.lua` in directory name.

Since the usage `-c config-xxx.lua` is supported and may be naturally used (by human) with the help of shell completion (type `-c config-` followed by a Tab), this PR unifies the name pattern of `testdir`. Now in both cases `testdir` without trailing `.lua` will be created.

------

Tip: To review changes, you may want to check "Hide whitespace" and reload
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ this project uses date-based 'snapshot' version identifiers.
 
 ## [Unreleased]
 
+### Changed
+- Unify `testdir` by dropping possibly trailing `.lua` passed to `--config`
+
 ### Fixed
 - Ensure directories `testdir` and `resultdir` exist when `--dirty` is set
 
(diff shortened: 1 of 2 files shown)


josephwright:
I'm not sure the intention was to support `-c <Lua-file-name>` rather than `-c <config-name>`, but I don't have an issue with the change. But what's the reason for the whitespace changes?

muzimuzhi:
> But what's the reason for the whitespace changes?

To use 2-space per indent level. Previously those lines were indented by 3 spaces. May be better illustrated with a screenshot, 

image

muzimuzhi:
> I'm not sure the intention was to support `-c <Lua-file-name>` rather than `-c <config-name>`

BTW, the `-c <config-name>.lua` support was added in 103fb51 (allow .lua suffix for -c option, 2019-02-01).

Issue #274, opened 2023-02-20
`ps2pdf` doesn't respect epoch settings


muzimuzhi:
When I was preparing #273 to restore the epoch settings in `dvitopdf()` and trying to test against this change, I find `ps2pdf` (or the ultimately called executable Ghostscript `gs`) doesn't respect epoch settings.

`dvitopdf()` executes `dvips` followed by `ps2pdf`. In ps files generated by `dvips`, timestampt is normalized. While in pdf files generated by `ps2pdf`, Ghostscript `gs` (v10.0.0) writes bunch of timestampts and version info in an XML stream and the `/Info` dictionary. Maybe `l3build` wants to normalize the whole XML stream or specific XML tags.

My branch [`epoch-gs-pdf`](https://github.com/muzimuzhi/l3build/tree/diff-gs-pdf) can be used as a start to check and test. Also see a typical diff in https://github.com/muzimuzhi/l3build/actions/runs/4225365084/jobs/7337547030
```xml
11 0 obj
<</Type/Metadata
/Subtype/XML/Length 1214>>stream
<?xpacket begin='﻿' id='W5M0MpCehiHzreSzNTczkc9d'?>
<?adobe-xap-filters esc="CRLF"?>
<x:xmpmeta xmlns:x='adobe:ns:meta/' x:xmptk='XMP toolkit 2.9.1-13, framework 1.6'>
<rdf:RDF xmlns:rdf='http://www.w3.org/1999/02/22-rdf-syntax-ns#' xmlns:iX='http://ns.adobe.com/iX/1.0/'>
<rdf:Description rdf:about="" xmlns:pdf='http://ns.adobe.com/pdf/1.3/' pdf:Producer='GPL Ghostscript 10.00.0'/>
<rdf:Description rdf:about="" xmlns:xmp='http://ns.adobe.com/xap/1.0/'><xmp:ModifyDate>2023-02-20T22:05:51+08:00</xmp:ModifyDate>
<xmp:CreateDate>2023-02-20T22:05:51+08:00</xmp:CreateDate>
<xmp:CreatorTool>dvips(k) 2022.1 (TeX Live 2022)  Copyright 2022 Radical Eye Software</xmp:CreatorTool></rdf:Description>
<rdf:Description rdf:about="" xmlns:xapMM='http://ns.adobe.com/xap/1.0/mm/' xapMM:DocumentID='uuid:DocumentUUID'/>
[...]

PR #278, opened 2023-02-22
Extend PDF-based tests to `dvips`

muzimuzhi:
This PR adds `dvips` chain to the list of check engines use for PDF-based tests, and of course the new test output.

With the help from @u-fischer in #274, timestamp metadata in PDFs generated by `ps2pdf` is normalized by adding `\DocumentMedadata{}` to tex file, which defines `\__pdf_backend_set_regression_data:` to be used by `regression-test.tex`.

Two by-produces (perhaps they should live in separate PRs, but due to relations with current PR...):
 - Add option `-a` to `diff` program call, to make it consider all files to be text files.
   - This is helpful when there're remaining byte streams in files to compare.
   - Right now I don't have direct access to Windows. From [doc](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/fc) it seems the corresponding option for `fc` in Windows is `/b`.
 - Add a new pattern of PDF stream to normalize PDFs.
   - In old version of test output `00-test-2.latexdvips.tpf`, two streams were not normalized.  Their beginning marks were `/Subtype/Type1C/Length 938>>stream` and `/Type/Metadata/Length 1546>>stream`, so I use a strict pattern `/Length %d+>>stream` to try to match them.  (a sample [00-test-2.latexdvips.tpf.zip](https://github.com/latex3/l3build/files/10806738/00-test-2.latexdvips.tpf.zip))

------

My original purpose was to test against #273, but it ends with findings that `dvips` chain (more specifically, the ghostscript behind `ps2pdf`) doesn't need the epoch settings in env vars to generate normalized metadata in PDF. Evidence: [checks are passed](https://github.com/muzimuzhi/l3build/actions/runs/4244240916) for commit af4950c which reverts #273.

The settings are kept, as it does no harm and is used by `dvips`.


CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ this project uses date-based 'snapshot' version identifiers.
 
 ## [Unreleased]
 
+### Added
+- Force non-Windows diff program to consider all files to be text files
+- Extend PDF-based tests to `dvips`
```
muzimuzhi, line 12: Does `l3build` test changes need to be listed in CHANGELOG?

build.lua
```diff
@@ -7,6 +7,16 @@ bundle = ""
...
+--[[ FIXME:
+  - setting specialformats in config-pdf.lua results in "binary" set to "latexdvips" (should be "latex")
+  - setting specialformats in build.lua disables specifying engine
+      l3build save -c config-pdf -e latexdvips 00-test-2
+]]
+specialformats = specialformats or {}
+specialformats["latex"] = specialformats["latex"] or
+  {
+    latexdvips = {binary = "latex", format = ""}
+  }
```
muzimuzhi, line 19: I guess the wired limitations are caused by some `l3build` problems, but for now let's just use a way that "works".

l3build-check.lua
```diff
@@ -551,6 +551,11 @@ local function normalize_pdf(content)
       binary = false
       stream = true
       stream_content = "stream" .. os_newline
+    elseif match(line, "/Length %d+>>stream$") then
```
muzimuzhi, line 554: Maybe `">>stream$"` is enough, but I chose a more strict pattern, mainly to not match a non-stream.

CHANGELOG.md
josephwright: I'm not sure what you mean - that the change may require `.tlg` updates?

.github/tl_packages
```diff
@@ -12,7 +12,11 @@ tex
 xetex
 # Requirements for the tests
 amsfonts
+dvips
 etex-pkg
+l3experimental
+latex-lab
+pdfmanagement-testphase
```
muzimuzhi, line 19: Will these three packages too experimental to cause less stable test output files?

CHANGELOG.md
muzimuzhi: the changes only involving "using l3build to test l3build itself"

muzimuzhi:
> Add a new pattern of PDF stream to normalize PDFs.
> * In old version of test output `00-test-2.latexdvips.tpf`, two streams were not normalized.  Their beginning marks were `/Subtype/Type1C/Length 938>>stream` and `/Type/Metadata/Length 1546>>stream`, so I use a strict pattern `/Length %d+>>stream` to try to match them.  (a sample [00-test-2.latexdvips.tpf.zip](https://github.com/latex3/l3build/files/10806738/00-test-2.latexdvips.tpf.zip))

Setting `ps2pdfopts = " -dCompressFonts=false "` will decompress the second stream starting with `/Type/Metadata/Length 1546>>stream`. But adding option `-dCompressFonts=false` has no effect on the embedded font stream starting with `/Subtype/Type1C/Length 938>>stream`.

At the moment I can't find a ghostscript cli option acts like the `-a` option to [`mutool clean`](https://mupdf.com/docs/manual-mutool-clean.html), which results in ASCII Hex encode binary streams.

Just for more real cases, here are all the different beginning patterns found in `*.tpf` files in [`pdfresrouce` repo, `testfiles-dvips` directory](https://github.com/latex3/pdfresources/tree/main/testfiles-dvips):
```postscript
/N 3/Length 8>>stream
/ProcSet [/PDF/ImageB/Text]>>/Length 170>>stream
/Resources<</ProcSet [/PDF/ImageB]>>/Length 107>>stream
/Resources<</ProcSet [/PDF/ImageB]>>/Length 108>>stream
/Resources<</ProcSet [/PDF]>>/Length 8>>stream
/Size[256]/Length 12>>stream
/Subtype /text#2fplain/Length 13>>stream
/Subtype/Type1C/Length 1006>>stream
/Subtype/XML/Length 1144>>stream
/Type/EmbeddedFile/Length 21>>stream
/Type/Metadata/Length 1547>>stream
/yyy(bla)/Length 89>>stream
<</Filter/FlateDecode/Length 172>>stream
>>>>/Length 71>>stream
```

muzimuzhi:
> **Update:** I'm inclined to add `ps2pdfopts = " -dCompressStreams=false "` and make the pattern more strict to only match streams for embedded fonts and images. (First half is done.)

I have a feeling that `l3build` should set  `-dCompressStreams=false` by default, when calling `ps2pdf` in `dvitopdf()`.  Because the expectation is, the pdf-based test outputs should be decompressed.

This can be done either by adding the option to variable `ps2pdfopts` which will be passed to `ps2pdf`, or by setting env var [`GS_OPTIONS`](https://ghostscript.readthedocs.io/en/latest/Use.html#gs-options).

On macOS, `ps2pdf` options will finally be passed to gs executable `gs` twice, hence I'm a bit tend to the `GS_OPTIONS` way.
```
$ grep -B 2 -- '-dBATCH' `which ps2pdfwr`
# We have to include the options twice because -I only takes effect if it
# appears before other options.
exec "$GS_EXECUTABLE" $OPTIONS -q -P- -dNOPAUSE -dBATCH -sDEVICE=pdfwrite -sstdout=%stderr "-sOutputFile=$outfile" $OPTIONS "$infile"
```

zauguin:
> * so I use a strict pattern `/Length %d+>>stream` to try to match them.  (a sample [00-test-2.latexdvips.tpf.zip](https://github.com/latex3/l3build/files/10806738/00-test-2.latexdvips.tpf.zip))

This pattern would match pretty generic streams and rely on implementation details regarding spacing in order not to remove something important. That sounds rather fragile. IMO normalization should only happen for data which is known not to be important.

muzimuzhi:
> > * so I use a strict pattern `/Length %d+>>stream` to try to match them.  (a sample [00-test-2.latexdvips.tpf.zip](https://github.com/latex3/l3build/files/10806738/00-test-2.latexdvips.tpf.zip))
> 
> This pattern would match pretty generic streams and rely on implementation details regarding spacing in order not to remove something important. That sounds rather fragile. IMO normalization should only happen for data which is known not to be important.

Yes I too realized the fragility of the new pattern in https://github.com/latex3/l3build/pull/278#issuecomment-1441325092. But then with the following three pieces of info, I tend to think the new pattern is not as fragile as we may think.
- The use of gs option `-dCompressStreams=false`.
- The fact that `normalize_pdf()` actually checks if the stream content contains "binary bytes" (roughly speaking), and will not mistakenly eat human readable stream. `normalized_pdf()` is first introduced in 64825daddadc.
    https://github.com/latex3/l3build/blob/26a593f2cdcbe928feca01090748ad2786ea38c2/l3build-check.lua#L531-L549
- The fact (from pdf reference v1.7, sec. 3.1.1 Character Set) that whether the string `stream` is on its own line is just a taste of PDF writers, hence provides no indication about whether the stream is compressed or not.

josephwright:
Where did we get to here?

muzimuzhi:
I think this PR can be closed for now. Maybe I'll do another try in the future.

The controversial part is, is the new stream match-and-drop logic added to [`l3build-check.lua`](https://github.com/latex3/l3build/pull/278/files#diff-593135098ccaf38b6e67ab4e0960713edbad21344be0fda56e5fc2d8840c7cdd) safe?

PR #281, opened 2023-02-25
Fully drop `stdconfig`

muzimuzhi:
The variable `stdconfig` was introduced in
- 9da3ee2 (Have stdconfig in checkconfigs table as-standard, 2017-12-07)
- 0004839 (Handle named tests with multiple configs (see #<span></span>15), 2018-02-21)

and then dropped in
- 9e6709d (Drop stdconfig as "build.lua" is now fixed, 2018-03-06)


muzimuzhi:
The code (without current PR) happens to work because after setting `checkconfigs`, in `l3build.lua` the main function
https://github.com/latex3/l3build/blob/26a593f2cdcbe928feca01090748ad2786ea38c2/l3build.lua#L227-L228
is called both for case `#checkconfigs == 1` and case `#checkconfigs == 0`. This behavior (or decision) is sort of reasonable because the default config is always loaded.

PR #285, opened 2023-03-01
Generalize normalization of ghostscript version

muzimuzhi:
- match two-digit gs major version, like 10.00.0
- match both "GhostScript" and "Ghostscript" (it seems lowercased one is used in recent gs releases)

The normalization of ghostscript version in PDF-based tests is introduced in
- commit 81859b9 (Normalise GhostScript version in PDF tests, 2022-04-14) and
- commit 563db82 (Mumble, 2022-04-14)

### DIFF
--- a/l3build-check.lua
+++ b/l3build-check.lua
@@ -556,7 +556,7 @@ local function normalize_pdf(content)
       not match(line,"^%%%%%+") then
       line = gsub(line,"%/ID( ?)%[<[^>]+><[^>]+>]","/ID%1[<ID-STRING><ID-STRING>]")
       line = gsub(line,"%/ID( ?)%[(%b())%2%]","/ID%1[<ID-STRING><ID-STRING>]")
-      line = gsub(line,"GhostScript %d%.%d+%.?%d*","GhostScript ...")
+      line = gsub(line,"Ghost[sS]cript %d+%.%d+%.?%d*","Ghostscript ...")
       new_content = new_content .. line .. os_newline
     end
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ this project uses date-based 'snapshot' version identifiers.
 
 ## [Unreleased]
 
+### Changed
+- Generalize normalization of ghostscript version in PDF-based tests
+
 ## [2023-02-26]
 
 ### Changed


l3build-check.lua
```diff
@@ -556,7 +556,7 @@ local function normalize_pdf(content)
       not match(line,"^%%%%%+") then
       line = gsub(line,"%/ID( ?)%[<[^>]+><[^>]+>]","/ID%1[<ID-STRING><ID-STRING>]")
       line = gsub(line,"%/ID( ?)%[(%b())%2%]","/ID%1[<ID-STRING><ID-STRING>]")
-      line = gsub(line,"GhostScript %d%.%d+%.?%d*","GhostScript ...")
+      line = gsub(line,"Ghost([sS])cript %d+%.%d+%.?%d*","Ghost%1cript ...")
```
josephwright, line 559: If we are generalising, shouldn't the output version here be fixed?

josephwright (changes requested):

l3build-check.lua
muzimuzhi: You mean always substituted with one of uppercase or lowercase s, but not both?

At first I do want to use only "Ghostscript ...", because it's used by recent gs, checked with gs 10.00.0 on my macOS, 9.55.0 on github actions ubuntu LTS 22.04, 9.56.1 in @u-fischer's `pdfresources` repo whose testfiles are generated on windows from my guess.  I also guess when you implemented it about one year ago, at that time gs did write uppercased S.

It's the fear of breaking existing testfiles that makes me step back.  One can also argue the case of S is not part of the gs version.

muzimuzhi: Who knows what the pattern in future versions of entry `/Producer(GPL Ghostscript 9.55.0)` (the one generated by gs on ubuntu LTS 22.04) will be changed to... Will `/Producer\(.*Ghost[sS]cript.*\)` be too general? A line by line matching is too limited.

Ah wait, Is the `/Producer` entry overwritten-able?

josephwright: I meant that we already have `GhostScript` and `Ghostscript`, so normalising those two as well as the version number would be good.

muzimuzhi: haha I was complicating things...

PR #290, opened 2023-03-07
Documentation improved

dbitouze:
I was a bit confused while reading a printed version of the `l3build`'s documentation. Indeed :

1. cross-references did not mention page numbers while some referenced objects (especially floating listings) were very far away,
1. listings, made floating since within `figure` environments, were called sometimes "Figure", sometimes "Listings",
1. the case of referenced objects was not homogeneous,
1. the naming of referenced objects was not homogeneous.

The improvement was made possible thanks to :

- the `varioref` package (inconvenience 1),
- the floating listings feature of the `listings` package, which are no longer in `figure` environments (inconveniences 1 and 2),
- the `cleveref` package (inconveniences 2, 3 and 4).

### DIFF
--- a/l3build.dtx
+++ b/l3build.dtx
@@ -621,42 +621,42 @@
 % Here, the |module| only is defined, and since it doesn't use |.ins| files so the variable \var{unpackfiles} is redefined to run |tex| on the |.dtx| files instead to generate the necessary |.sty| files.
 % There are some PDFs in the repository that shouldn't be part of a CTAN submission, so they're explicitly excluded, and here unpacking is done `quietly' to minimise console output when building the package.
 %
-%   \begin{lstlisting}[frame=single,language={[5.2]Lua},gobble = 6,float,label=fig:breqn,caption=The build configuration for the \pkg{breqn} package.]
-%     -- Build configuration for breqn
+% \begin{floating-listing}[caption=The build configuration for the \pkg{breqn} package.,label=lst:breqn]
+%   -- Build configuration for breqn
 %
-%     module = "breqn"
+%   module = "breqn"
 %
-%     unpackfiles = {"*.dtx"}
-%     excludefiles = {"*/breqn-abbr-test.pdf",
-%                     "*/eqbreaks.pdf"}
-%     unpackopts  = "-interaction=batchmode"
-%   \end{lstlisting}
+%   unpackfiles = {"*.dtx"}
+%   excludefiles = {"*/breqn-abbr-test.pdf",
+%                   "*/eqbreaks.pdf"}
+%   unpackopts  = "-interaction=batchmode"
+% \end{floating-listing}
 %
 % An example of a bundle build script for \pkg{l3packages} is shown in Listing~\vref{lst:bundle}.
 % Note for \LaTeX{} we use a common file to set all build variables in one place, and the path to the |l3build.lua| script is hard-coded so we always use our own most recent version of the script.
 % An example of an accompanying module build script is shown in Listing~\vref{lst:module}.
 %
-%   \begin{lstlisting}[frame=single,language={[5.2]Lua},gobble = 6,float,caption={The build script for the \pkg{l3packages} bundle.},label={fig:bundle}]
-%     -- Build script for LaTeX "l3packages" files
+% \begin{floating-listing}[caption={The build script for the \pkg{l3packages} bundle.},label={lst:bundle}]
+%   -- Build script for LaTeX "l3packages" files
 %
-%     -- Identify the bundle: there is no module as this is the "driver"
-%     bundle = "l3packages"
+%   -- Identify the bundle: there is no module as this is the "driver"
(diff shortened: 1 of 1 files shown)


josephwright:
I'd like to see what others think here. I've no issue with `varioref`, other than one could argue that it makes the sources slightly less clear. The `listings` part I worry about semantics: you loose the source separation of 'a floating environment' and 'a verbatim-like code input'. In the past I've had `fun` with `cleveref`, but perhaps that is sorted re. `hyperref` (@u-fischer?).

dbitouze:
> I've no issue with `varioref`, other than one could argue that it makes the sources slightly less clear.

I guess you meant `cleveref` here since, concerning `varioref`, `\vref{...}` isn't less clear than `\ref{...}`.

> The `listings` part I worry about semantics: you loose the source separation of 'a floating environment' and 'a verbatim-like code input'.

Is it really that important? The first concern should be the readability of the document read by the target readers (I mean the output).

> In the past I've had `fun` with `cleveref`, but perhaps that is sorted re. `hyperref` (@u-fischer?).

Works pretty well since years for me.

josephwright:
> > I've no issue with `varioref`, other than one could argue that it makes the sources slightly less clear.
> 
> I guess you meant `cleveref` here since, concerning `varioref`, `\vref{...}` isn't less clear than `\ref{...}`.

Nope, I meant `varioref`: compare `Figure~\ref{fgr:some-stuff}` with `\vref{fgr:some-stuff}` - it's much easier to read the first version in the source.

> > The `listings` part I worry about semantics: you loose the source separation of 'a floating environment' and 'a verbatim-like code input'.
> 
> Is it really that important? The first concern should be the readability of the document read by the target readers (I mean the output).

As the team are working on tagging, I'd say yes. What I'm suggesting is we should define a dedicated code-listing float type for the docs, then nest the `listing` or `verbatim` or whatever inside it.

u-fischer:
> In the past I've had `fun` with `cleveref`, but perhaps that is sorted re. `hyperref`

I wouldn't use cleveref for our documentation. I'm not aware of a special current problem with hyperref but cleveref redefines core commands like `\refstepcounter` and `\label` and makes use of internal hyperref command and that could get problematic if at some time we want to add tagging to our documentation: the maintainer seems to be not active, and the code is quite complicated and difficult to adjust from the outside.

josephwright:
@u-fischer Thanks: aligns with my impression from the past. I agree that means we should avoid `cleveref`.

dbitouze:
> > I guess you meant `cleveref` here since, concerning `varioref`, `\vref{...}` isn't less clear than `\ref{...}`.
> 
> Nope, I meant `varioref`: compare `Figure~\ref{fgr:some-stuff}` with `\vref{fgr:some-stuff}` - it's much easier to read the first version in the source.

I insist :wink:: that's `cleveref` and not `varioref` that lets us write `\vref{fgr:some-stuff}` instead of `Figure~\ref{fgr:some-stuff}`.

dbitouze:
> > In the past I've had `fun` with `cleveref`, but perhaps that is sorted re. `hyperref`
> 
> I wouldn't use cleveref for our documentation. I'm not aware of a special current problem with hyperref but cleveref redefines core commands like `\refstepcounter` and `\label` and makes use of internal hyperref command and that could get problematic if at some time we want to add tagging to our documentation: the maintainer seems to be not active, and the code is quite complicated and difficult to adjust from the outside.

OK, I'll adjust my PR. Too bad there isn't a new, cleaner, `cleveref`: its features are very useful!

u-fischer:
> Too bad there isn't a new, cleaner, cleveref: its features are very useful!

you could take a look at zref-clever. But I wouldn't use it for the documentation either currently. Less external packages will make a first tagging easier.

FrankMittelbach:
> > Too bad there isn't a new, cleaner, cleveref: its features are very useful!
> 
> you could take a look at zref-clever. But I wouldn't use it for the documentation either currently. Less external packages will make a first tagging easier.

I think we should refrain from adding cleverref or zref-clever to our documentation but try to keep the dependencies on thrid-party packages as low as possible (unless it is absolutely essential).

FrankMittelbach:
> Nope, I meant `varioref`: compare `Figure~\ref{fgr:some-stuff}` with `\vref{fgr:some-stuff}` - it's much easier to read the first version in the source.

nope :-) @dbitouze is correct, that is not varioref but cleverref that allows the above. varioref just adds an additional page reference as appropriate.

dbitouze:
> I think we should refrain from adding cleverref or zref-clever to our documentation but try to keep the dependencies on third-party packages as low as possible (unless it is absolutely essential).

Hopefully, you don't suffer from schizophrenia so you don't consider `varioref` as a third party package :wink :

josephwright:
Forgot to come back on this. Looks good except the 'local variables' bit: I'll merge then tidy up.

PR #299, opened 2023-05-19
Clarify `\TEST` and `\TESTEXP` docs

muzimuzhi:
To help myself when I forgot if `\TEST` runs in a group again ...

- Before
  image
- After
  image


muzimuzhi:
Oops I forgot to commit and push local change to changelog.  Good news is that there were already unreleased changelog entries.

Issue #303, opened 2023-07-16
Refine error mesasage "attempt to index a nil value"


koppor:
I am a newcomer to l3build. Wanted to "quickly" create a GitHub actions based CI environment for [tagpdf](https://github.com/u-fischer/tagpdf). However, I cannot get `l3build check` working on the repository. The output is "attempt to index a nil value".

Is it possible to output a more informative error message?

```
$ l3build check
Running l3build with target "check" with configuration build
C:/MiKTeX/scripts/l3build/l3build-unpack.lua:87: attempt to index a nil value
Running l3build with target "check" with configuration config-pdftex
C:/MiKTeX/scripts/l3build/l3build-unpack.lua:87: attempt to index a nil value
Running l3build with target "check" with configuration config-luatex
C:/MiKTeX/scripts/l3build/l3build-unpack.lua:87: attempt to index a nil value
Failed tests for configuration build:

  Check failed with difference files

Failed tests for configuration config-pdftex:

  Check failed with difference files

Failed tests for configuration config-luatex:

  Check failed with difference files
```

The respective part of `l3build-unpack.lua` looks like follows in my MiKTeX installation?

```
      local success = io.popen(
        "cd " .. unpackdir .. "/" .. path .. os_concat ..
        os_setenv .. " TEXINPUTS=." .. os_pathsep
          .. localdir .. (unpacksearch and os_pathsep or "") ..
        os_concat  ..
        os_setenv .. " LUAINPUTS=." .. os_pathsep
          .. localdir .. (unpacksearch and os_pathsep or "") ..
        os_concat ..
        unpackexe .. " " .. unpackopts .. " " .. name
          .. (options["quiet"] and (" > " .. os_null) or ""),
        "w"
      ):write(string.rep("y\n", 300)):close()
```

josephwright:
@koppor  Ah, MiKTeX :) @u-fischer mentioned that we should look at this again (basically, it doesn't work as MiKTeX doesn't return the same info as TeX Live). I promised I'd look at her issue on my way back, so I'll add this to the immediate list.

josephwright:
It's not out fault :) You can minimise to 
```lua
--[[

% !TeX program = texlua

--]]

-- io.popen("echo","w"):write(""):close() -- Works
-- io.popen("echo","w"):write("")         -- Works
io.popen("echo","w"):write("y"):close()
```
[...]

PR #305, opened 2023-07-16
Add MiKTeX check

koppor:
To enable working on https://github.com/latex3/l3build/issues/303, this PR adds MiKTeX as separated check.

Output with the check:

```
Run texlua l3build.lua ctan -H --show-log-on-error
Running l3build with target "check" for module "."
Running l3build with target "check" and configuration "build"
./l3build-unpack.lua:87: attempt to index a nil value
```

Hope, this will make it easier to work on a fix.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -23,3 +23,23 @@ jobs:
           package_file: .github/tl_packages
       - name: Run l3build
         run: texlua l3build.lua ctan -H --show-log-on-error
+  miktex:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Install MikTeX
+        run: |
+          set -e
+          sudo gpg --homedir /tmp --no-default-keyring --keyring /usr/share/keyrings/miktex.gpg --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys D6BC243565B2087BC3F897C9277A7293F59E4889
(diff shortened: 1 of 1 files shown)


josephwright:
I'm a bit worried about running a MiKTeX check on anything other than Windows: seems rather odd to me! (Yes, I know that CS has extended MiKTeX to other platforms, but Windows is the place it really used.)

koppor:
It appears "strange" to me too. I see, however, that the debian packages are kept being updated on CTAN: https://ctan.net/systems/win32/miktex/setup/deb/dists/

image

Since there are many Windows users out there - an automatic check would be beneficial.

I got two proposals to proceed. Just let me know.

Proposal A (more work for me):

Port this PR to Windows. Meaning: Use the GitHub action Windows runner, install MiKTeX for Windows and let the thing run on Windows.

Proposal B (more risk for you):

- Fix https://github.com/latex3/l3build/issues/303
- Merge that fix into this branch
- Check if branch runs through
- Merge the branch into `main`
- In case the CI works broken l3build, but MiKTeX on Windows does not, remove this check and implement proposal A.

josephwright:
I think before we tackle either route we need to see if MiKTeX is workable: like I've said in #303, there is an issue with allowable rounding differences.

koppor:
Then, I close this, because it will take longer time.

The "allowable rounding differences" will be a hard issue. Maybe, testing with l3build will be recommended on TeXLive only.

josephwright:
@koppor I'll see what I get back from Christian: that's a blocker for us on MiKTeX, unfortunately.

PR #323, opened 2023-12-07
Document syntax of `\SHOWFILE` and `\ASSERT(STR)`

muzimuzhi:
I think syntaxes are preferable than textual descriptions.

Before
image
After
image


muzimuzhi:
Hmm looks like `<file name>` is preferred than `<filename>`.

In `l3kernel` repository,
```
$ grep -rn '\\\(Arg\|meta\){filename}' **/*.dtx | wc -l
       3
$ grep -rn '\\\(Arg\|meta\){file name}' **/*.dtx | wc -l
      47
```

muzimuzhi:
Coincidentally, `l3build.dtx` uses "filename" only. I'll leave it as is.

PR #333, opened 2024-01-02
Get numeric error-level from `os.execute`

muzimuzhi:
This PR makes sure `l3build` gets an numeric error-level from `os.execute()`.

**Background**

Till Lua 5.1.x, `os.execute()` returns only one result which represents status code.

Starting from Lua 5.2, `os.execute()` returns 3 results, among which the first result is either `true` or `nil`, thus it can no longer be used as numeric `errorlevel` in l3build.

See Lua 5.2 Reference Manual, [8.2 – Changes in the Libraries](https://www.lua.org/manual/5.2/manual.html#8.2)

> Function `os.execute` now returns `true` when command terminates successfully and `nil` plus error information otherwise.

Full doc of [`os.execute()`](https://www.lua.org/manual/5.3/manual.html#pdf-os.execute) in Lua 5.3, the Lua version that latest LuaTeX (hence `texlua`) uses.

> `os.execute ([command])`
> This function is equivalent to the ISO C function system. It passes command to be executed by an operating system shell. Its first result is `true` if the command terminated successfully, or `nil` otherwise. After this first result the function returns a string plus a number, as follows:
>
> - "`exit`": the command terminated normally; the following number is the exit status of the command.
> - "`signal`": the command was terminated by a signal; the following number is the signal that terminated the command.
When called without a command, os.execute returns a boolean that is true if a shell is available.

### DIFF
--- a/build.lua
+++ b/build.lua
@@ -93,7 +93,7 @@ function update_tag(file,content,tagname,tagdate)
 end
 
 function tag_hook(tagname)
-  os.execute('git commit -a -m "Step release tag"')
+  return select(3, os.execute('git commit -a -m "Step release tag"'))
 end
 
 if not release_date then
--- a/l3build-check.lua
+++ b/l3build-check.lua
@@ -695,7 +695,7 @@ function base_compare(test_type,name,engine,cleanup)
   if compare then
     return compare(difffile, reffile, genfile, cleanup, name, engine)
   end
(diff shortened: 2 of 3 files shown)


muzimuzhi:
Oh, `os.execute()` from `texlua` still returns a single, numeric result.

I found a related mailing list thread https://tug.org/pipermail/luatex/2015-November/005536.html .

PR #340, opened 2024-01-05
Stick to third-person singular form in target and option docs

muzimuzhi:


car222222 (commented):
Three(?) places “and exit.” should become “and exits.”
Similarly, “simply rerun tests” probably should become “… reruns …”

Aside: I cannot parse this sentence:  … to … to … ??  — 
Applies the Lua |update_tag()| function to modify the contents of files specified by |tagfiles| to update the `release tag' (or package version) and date.

Aside: Should this have been: … run, even if … — 
Forces tests to run if engine is not set up

muzimuzhi:
Hmm the ["Co-authored-by" trailers](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-the-command-line) in git commit messages don't work as expected.

josephwright:
We good to go here?

FrankMittelbach (approved):
looks fine to me (but what do I know :-) )

car222222 (approved):
Apologies for the delay. 

I am not familiar with “assertions”, or the way that they are normally defined or described.  
But the ”assert” here looks strange here!  “true if . . . “ would be normal English:  
assert if the full expansion of \meta{arg_1} and \meta{arg_2} are the same:

car222222 (changes requested):
Lines 594-6. 
I think that the following is much clearer (if accurate:-) :

Applies the Lua |update_tag()| function to modify the contents of 
all the filesspecified by |tagfiles|:
this function updates the `release tag' (or package version)
and date.

car222222:
@josephwright (I guess).

The major problem with the \ASSERT item in this list is that it does not say anything about what actually gets written to the log file.  

Thus it seems to need a total rewrite.  
Or am I missing something?

josephwright:
@car222222  Well perhaps but that's a separate issue - suggest we agree on the PR here then I re-work that sentence later.

car222222:
@josephwright Agreed that this is something else again!  
Therefore I did approve this PR.

car222222:
My comment on Lines 594-6 got mangled a bit, try this instead.  

Applies the Lua |update_tag()| function 
to modify the contents of
all the files specified by |tagfiles|;
this function updates the `release tag' 
(or package version) and the date.

josephwright:
> Well perhaps but that's a separate issue - suggest we agree on the PR here then I re-work that sentence later.

@car222222 3e8067ea3901b5db759f06f46a0eb411331a7568

car222222:
I think that a space got lost in 0231d16, should be "; this". Probably my fault.

PR #346, opened 2024-01-17
Add switch -s|--stdengine

josephwright:


muzimuzhi:
Need to be added to the big list of options
https://github.com/latex3/l3build/blob/3d51aaea98479f7afeb5ae6e66d382106232c27f/l3build.dtx#L369-L372

Also searching for `-s` in `l3build.pdf` I spot an existing typo mentioning `-s` (lowercase s) (on line 460, see below), which was introduced in c4471ca (Add new --rerun|-r option (#<span></span>17), 2017-07-01) hence looks like it should be `-r` (short for `--rerun`).
https://github.com/latex3/l3build/blob/3d51aaea98479f7afeb5ae6e66d382106232c27f/l3build.dtx#L454-L461

l3build-check.lua
```diff
@@ -595,7 +595,9 @@ function runcheck(name, hide)
     return 1
   end
   local checkengines = checkengines
-  if options["engine"] then
+  if options["stdengine"] then
+    checkengines = {stdengine}
+  elseif options["engine"] then
     checkengines = options["engine"]
```
FrankMittelbach, line 601: not really for this pr but seeing this line: shouldn't in case of -e the checkengines variable set to the \cap of "engine" and the existing "checkengines" of a config file?

l3build.1
```diff
@@ -77,6 +77,8 @@ Shows the full log of the failure with 'halt-on-error'
 Shows the invocation to update failing .tlg files
 .IP --shuffle
 Shuffles order of tests
+.IP --stdengine|-s
+Run tests only with the standard engine
```
FrankMittelbach, line 81: I think it wouldn't hurt if one mentions in places like this that this can differ from config file to config file in a complex setup

FrankMittelbach (approved):
somebody remarked that this should be also in the main (pdf) documentation. Otherwise, looks fine.

l3build-check.lua
josephwright: Not necessarily - you can add `-f` for force running with an 'unsuitable' engine

josephwright:
```lua
-- Sanity check
function check_engines(config)
  if options["engine"] and not options["force"] then
    -- Make a lookup table
    local t = { }
    for _, engine in pairs(checkengines) do
      t[engine] = true
    end
    for _, engine in pairs(options["engine"]) do
      if not t[engine] then
        print("\n! Error: Engine \"" .. engine .. "\" not set up for testing with configuration \"" .. config .. "\"!")
        print("\n  Valid values are:")
        for _, engine in ipairs(checkengines) do
          print("  - " .. engine)
        end
        print("")
        exit(1)
      end
    end
  end
end
```

FrankMittelbach: I nevery really understood the use case for force :-) but yes, that makes it a little more complicated. But in case of asking for `-e`without `-f`: wouldn't it be more oportune to simply not do any tests for any of the "e" engines that are not in "checkengines" of a particular config rather calling it an error? Well maybe not, dunno

josephwright: Well we could drop `--force`: it was there for some setup stuff which likely we no longer need (and I think `-s` makes redundant).

josephwright: That could be done separate from the PR of course

FrankMittelbach: we could I guess, and yes it should be done in a separate pr.

l3build.dtx
```diff
@@ -395,6 +395,8 @@
 %   in the full \texttt{.log} file of a failed test to be shown on the console
 % \item |--show-saves| (|-S|) When tests fail, prints the \texttt{l3build save} commands needed
 %   to regenerate the tests assuming that the failures were false negatives.
+% \item |--stdengine| (|-s|) Run tests only with the standard engine (which can vary between
+%   configurations)
 % \item |--shuffle| Shuffles the order in which tests run
```
muzimuzhi, line 400: 't' comes after 'h'. (Entries in `l3build.1` is already ordered.)
```suggestion
% \item |--shuffle| Shuffles the order in which tests run
% \item |--stdengine| (|-s|) Run tests only with the standard engine (which can vary between
%   configurations)
```

muzimuzhi (changes requested):

PR #347, opened 2024-01-18
Drop switch --force|-f

josephwright:
Unknown engines are then skipped with a message, rather than leading to an error.

### DIFF
--- a/l3build-arguments.lua
+++ b/l3build-arguments.lua
@@ -307,4 +307,5 @@ function check_engines(config)
       end
     end
   end
+  if not next(checkengines) then exit(0) end
 end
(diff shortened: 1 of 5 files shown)


josephwright:
@muzimuzhi Would welcome your thoughts here (I can't directly add you as a reviewer).

l3build-arguments.lua
```diff
@@ -298,22 +292,20 @@ options = argparse()
...
+    checkengines = {}
+    for _,engine in ipairs(options["engine"]) do
+      if t[engine] then
+          insert(checkengines,engine)
+      else
+        print("Skipping unknown engine " .. engine)
       end
     end
   end
+  if not next(checkengines) then exit(0) end
```
FrankMittelbach, line 310: not sure it helps or not, but perhaps add a print("No applicable engine requested, config ignored")

FrankMittelbach (approved):
Assuming that force isn't used, yes that is what I had in mind

PR #355, opened 2024-06-20
doc: |options| and some functions

jlaurens:
documentation improvements: consistency and precision consistent wording in ...aux.lua
*fmt in .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ build/*
 *.log
 *.out
 *.toc
+*.fmt
--- a/l3build.dtx
+++ b/l3build.dtx
@@ -1671,25 +1671,25 @@
 %     \toprule
 %     Entry & Type \\
 %     \midrule
-%       \var{config}        & table   \\
-%       \var{date}          & string  \\
-%       \var{dirty}         & boolean \\
-%       \var{dry-run}       & boolean \\
-%       \var{email}         & string  \\
-%       \var{engine}        & table   \\
-%       \var{epoch}         & string  \\
-%       \var{file}          & string  \\
-%       \var{first}         & boolean \\
-%       \var{full}          & boolean \\
-%       \var{halt-on-error} & boolean \\
-%       \var{help}          & boolean \\
-%       \var{message}       & string  \\
-%       \var{names}         & table \\
-%       \var{quiet}         & boolean \\
-%       \var{rerun}         & boolean \\
-%       \var{shuffle}       & boolean \\
-%       \var{stdengine}     & boolean \\
-%       \var{texmfhome}     & string  \\
+%       \var{config}        & |table|   \\
+%       \var{date}          & |string|  \\
+%       \var{dirty}         & |boolean| \\
+%       \var{dry-run}       & |boolean| \\
+%       \var{email}         & |string|  \\
+%       \var{engine}        & |table|   \\
+%       \var{epoch}         & |string|  \\
+%       \var{file}          & |string|  \\
+%       \var{first}         & |boolean| \\
+%       \var{full}          & |boolean| \\
+%       \var{halt-on-error} & |boolean| \\
+%       \var{help}          & |boolean| \\
+%       \var{message}       & |string|  \\
+%       \var{names}         & |table|   \\
+%       \var{quiet}         & |boolean| \\
+%       \var{rerun}         & |boolean| \\
+%       \var{shuffle}       & |boolean| \\
+%       \var{stdengine}     & |boolean| \\
+%       \var{texmfhome}     & |string|  \\
 %     \bottomrule
 %     \end{tabular}
 %   \end{center}
--- a/l3build.dtx
+++ b/l3build.dtx
@@ -335,12 +335,13 @@
 % The \texttt{build.lua} file used to control \pkg{l3build} is a simple Lua file
 % which is read during execution. In the current release of \pkg{l3build},
 % \texttt{build.lua} is read automatically and can access all of the global
-% functions provided by the script. Thus it may contain a simple list of
+% functions and variables provided by the script. Thus it may contain a simple list of
 % variable settings \emph{or} additional code to customize the build
 % process.
 %
 %The example scripts given in Section~\vref{sec:examples} largely cover the required knowledge in Lua programing.
 % For a more advanced usage, one may consult general Lua documentations including \url{https://www.lua.org/manual/5.3/manual.html} and for the few |texlua| specific additions see section 4.2 of the \LuaTeX{} manual available locally with |texdoc luatex| command line or at \url{https://www.pragma-ade.com/general/manuals/luatex.pdf}.
+% Note that |os.execute| is at version 5.1, which is undocumented there.
 %
 % \subsection{Main build targets}
 %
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -19,3 +19,4 @@ build/*
 *.log
 *.out
 *.toc
+*.fmt
```
josephwright, line 22: Why this change?

l3build.dtx
```diff
@@ -335,12 +335,13 @@
...
 % which is read during execution. In the current release of \pkg{l3build},
 % \texttt{build.lua} is read automatically and can access all of the global
-% functions provided by the script. Thus it may contain a simple list of
+% functions and variables provided by the script. Thus it may contain a simple list of
 % variable settings \emph{or} additional code to customize the build
 % process.
 %
 %The example scripts given in Section~\vref{sec:examples} largely cover the required knowledge in Lua programing.
 % For a more advanced usage, one may consult general Lua documentations including \url{https://www.lua.org/manual/5.3/manual.html} and for the few |texlua| specific additions see section 4.2 of the \LuaTeX{} manual available locally with |texdoc luatex| command line or at \url{https://www.pragma-ade.com/general/manuals/luatex.pdf}.
+% Note that |os.execute| is at version 5.1, which is undocumented there.
```
josephwright, line 344: I don't follow the need for this - LuaTeX includes Lua 5.3.

l3build.dtx
```diff
@@ -368,15 +369,15 @@
 %
 % As well as these targets, the system recognises the options
 % \begin{itemize}
-% \item |--config| (|-c|) Configuration(s) to use for testing
+% \item |--config| (|-c|) Comma or space separated list of configurations to use for testing
```
josephwright, line 372: The comma/space business is simply standard command line rules based on `--XXX` or `-X` options - I don't see that we need to mention this specifically.

l3build.dtx
```diff
@@ -388,14 +389,14 @@
...
-% \item |--show-log-on-error| To be used in addition to \texttt{--halt-on-error} and results
+% \item |--show-log-on-error| When in addition to \texttt{-{}-halt-on-error}, results
 %   in the full \texttt{.log} file of a failed test to be shown on the console
-% \item |--show-saves| (|-S|) When tests fail, prints the \texttt{l3build save} commands needed
-%   to regenerate the tests assuming that the failures were false negatives.
+% \item |--show-saves| (|-S|) When tests fail, prints the \texttt{l3build save ...} commands needed
+%   to regenerate the tests, assuming that the failures were false negatives.
 % \item |--shuffle| Shuffles the order in which tests run
-% \item |--stdengine| (|-s|) Run tests only with the standard engine (which can vary between
+% \item |--stdengine| (|-s|) Runs tests only with the standard engine (which can vary between
```
josephwright, line 397: Deliberate choice

l3build.dtx
```diff
@@ -1662,35 +1663,40 @@
...
-%       \var{rerun}         & Boolean \\
-%       \var{shuffle}       & Boolean \\
-%       \var{stdengine}     & Boolean \\
-%       \var{texmfhome}     & String  \\
+%       \var{names}         & table \\
+%       \var{quiet}         & boolean \\
+%       \var{rerun}         & boolean \\
+%       \var{shuffle}       & boolean \\
+%       \var{stdengine}     & boolean \\
+%       \var{texmfhome}     & string  \\
```
josephwright, line 1692: These are deliberately capitalised as the cell starts a statement

l3build.dtx
```diff
@@ -1662,35 +1663,40 @@
...
 %     \bottomrule
 %     \end{tabular}
 %   \end{center}
+%   The Lua tables mentionned here are in fact arrays of strings.
+%   From the |build.lua| file, one can modify the string and boolean values,
+%   add or remove entries in arrays. But it is not recommended to affect a
+%   whole new Lua table to |options| nor to its array entries.
```
josephwright, line 1699: I',m not sure what you are getting at: `options` is set by `l3build <options>`

l3build.dtx
```diff
@@ -1887,32 +1893,32 @@
 %
 % \subsection{Components of \texttt{l3build}}
 %
+% Next functions return $0$ on a successful completion, a non $0$ error code otherwise.
```
josephwright, line 1896: I'm not sure what this is referring to here in 'next'

josephwright (changes requested):

l3build.dtx
jlaurens: This is the only one, why the other "Runs", "Instructs", "Prints"?

l3build.dtx
jlaurens: Here the type cannot really be capitalized because it refers to 2 things
1) the output of the lua function `type`
2) section 5.6 "Each entry needs a type one of boolean, string or table.
So these are not words, these are case sensitive keywords that cannot be capitalized.
In fact they should be formatted as in section 5.6: I'll commit a change.

l3build.dtx
jlaurens: Apparently, options is not a read only variable. It is set by `l3buil-arguments` before `build.lua` is executed and modified afterwards. Nothing prevents people from saying
```
options={...}
```
or
```
options.engine = {...}
```
in `build.lua` or `config-....lua`. This is an open door to insidious bugs.

l3build.dtx
jlaurens: The functions documented in the section?

I do not really understand the title of the section, as compared to the following section.

l3build.dtx
jlaurens: Is this section 5.6 still useful?

l3build.dtx
jlaurens: Except for `os.execute` that is not at version 5.3 but 5.1.
See [https://tug.org/pipermail/luatex/2015-November/005536.htm](https://tug.org/pipermail/luatex/2015-November/005536.html)

l3build.dtx
jlaurens: This is standard for people who know the standard. The cost of the precision is epsilon.

jlaurens:
Last commit: In the documentation of |call|, there was
```
The ⟨options⟩ table should take the same form as the global ⟨options⟩, described above.[...].
Note that any entry for the |target| in this table is ignored.
```
Not |target| field was exposed in the |options| table.

l3build.dtx
josephwright: I mean though that this applies to every command line tool one might use, more or less - e.g. `git` docs don't say this sort of thing

josephwright: BTW, it's not comma-or-space, it's just comma - spaces are used to separate arguments

jlaurens: This is not what is implemented.
```
l3build check -e "foo bar"
```
this needs clarification either in the implementation of in the documentation

jlaurens: [Lines 270-272](https://github.com/latex3/l3build/blob/5db0d64ba5647741114e3f0a7002a85fd1775bdd/l3build-arguments.lua#L270) of the `argparse` function reads
```
          for hit in gmatch(optarg, "([^,%s]+)") do
            insert(opts, hit)
          end
```
This is where argument values with spaces are managed. If spaces are not supported then the `%s` is misleading and should be removed.

josephwright: This is there to support the `--<long>[=]<option>` syntax in addition to the `-<short><option>` - you can quickly test that `--long option option2` treats `option2` as a value not an option.

jlaurens: Not really,
* `--<long>=<option>` is managed at line 228
* `--<long> <option>` is managed at line 253
* `--<short><option>` is managed at line 237
id est before line 270. This line is just there to split `<option>` into a list. The separator is anything that contains a comma _or a space_.

josephwright: It was a while ago .. I will check :)

josephwright: Looking back, I used `,%s` since I added `argparse`-like syntax at all - I guess just an oversight

josephwright: I've sorted this - just commas count here

josephwright:
I think it would be useful if you could split the PR into ones that deal with just one section/concept each - some are easy to agree to, others as we've seen more open ...

PR #361, opened 2024-06-26
UTC date in zip archive

jlaurens:
avoid local date in tagging.
It assumes that the `os.date` is not buggy.
Fix a bug in l3build-zip

### DIFF
--- a/l3build-zip.lua
+++ b/l3build-zip.lua
@@ -22,18 +22,19 @@ for those people who are interested.
 
 --]]
 
-local concat = table.concat
-local open = io.open
-local osdate = os.date
-local pack = string.pack
+local concat  = table.concat
+local open    = io.open
+local os_date = os.date
+local os_time = os.time
+local pack    = string.pack
 local setmetatable = setmetatable
-local iotype = io.type
+local iotype  = io.type
 
 local compress = zlib.compress
 local crc32 = zlib.crc32
 
 local function encode_time(unix)
-  local t = osdate('*t', unix)
+  local t = os_date('!*t', unix)
   local date = t.day | (t.month << 5) | ((t.year-1980) << 9)
   local time = (t.sec//2) | (t.min << 5) | (t.hour << 11)
   return date, time
@@ -75,7 +76,7 @@ local meta = {__index = {
     if #compressed >= #content then
       compressed = nil
     end
-    local timestamp = os.time()
+    local timestamp = os_time()
     local date, time = encode_time(timestamp)
     local central_extra, local_extra = extra_timestamp(timestamp, nil, nil)
     z.f:write(pack("<c4I2I2I2I2I2I4I4I4I2I2",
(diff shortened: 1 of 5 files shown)


josephwright:
I'm not sure this makes sense: if the package date is ahead of the local date when you build docs, it could well fail.

jlaurens:
What is the link between the `doc` and the date?

josephwright:
For example when `l3doc` checks that things are not deprecated ...

jlaurens:
Comparing times (or dates) in possibly different time zones gives undefined result, if you don't know the time zones in advance. `l3doc` _silently_ assumes that the dates are in the local TZ, this should be documented in `l3doc`.
New commit with changes replaced by documentation.

l3build-zip.lua
```diff
@@ -22,18 +22,19 @@ for those people who are interested.
...
-local osdate = os.date
-local pack = string.pack
+local concat  = table.concat
+local open    = io.open
+local os_date = os.date
+local os_time = os.time
+local pack    = string.pack
 local setmetatable = setmetatable
-local iotype = io.type
+local iotype  = io.type
```
josephwright, line 31: Not really seeing the need for this but at the same time I have no strong objection

l3build-zip.lua
```diff
@@ -22,18 +22,19 @@ for those people who are interested.
...
 
 local compress = zlib.compress
 local crc32 = zlib.crc32
 
 local function encode_time(unix)
-  local t = osdate('*t', unix)
+  local t = os_date('!*t', unix)
```
josephwright, line 37: This one is a job of @zauguin

josephwright (approved):
I'm happy with the doc change - but I'd like to see what Marcel thinks before moving forward, as the zip code is his.

zauguin:
Not sure why this is closed, but I don't have strong opinions on this. I'm weakly in favor of the change, but I'm guessing we should ask CTAN what they usually assume the times to mean in uploads.

jlaurens:
The [libzip documentation](https://libzip.org/specifications/extrafld.txt) is very clear on the subject.

zauguin:
> The [libzip documentation](https://libzip.org/specifications/extrafld.txt) is very clear on the subject.

The libzip file you linked only documents `extra` fields. I'm guessing you are referring to the `UT` (extra timestamp) extra field which in fact uses UTC based timestamps and which we generate. These timestamps are already correctly UTC encoded.

The function you propose to change is for the legacy timestamps. Their specification in APPNOTE says:

> The date and time are encoded in standard MS-DOS format.
> If input came from standard input, the date and time are
> those at which compression was started for this data. 
> If encrypting the central directory and general purpose bit 
> flag 13 is set indicating masking, the value stored in the 
> Local Header will be zero. MS-DOS time format is different
> from more commonly used computer time formats such as 
> UTC. For example, MS-DOS uses year values relative to 1980
> and 2 second precision.

I wouldn't refer to UTC as a "time format" but in any case it explicitly does not specify UTC for these. Also MS-DOS times are traditionally in the local timezone.
Since you referenced libzip, [libzip uses the local timezone for these](https://github.com/nih-at/libzip/blob/aa90b70e552709316cd2144837c3dd13b5fa1ec3/lib/zip_dirent.c#L1117).

jlaurens:
Thanks for the clarification.

PR #365, opened 2024-06-27
Doc: Type -> Lua type

jlaurens:
Description of the `options` table:

1. some types were not capitalized
2. all types are now Lua types, conform to section 5.6
3. |target| is exposed

### DIFF
--- a/l3build.dtx
+++ b/l3build.dtx
@@ -1667,27 +1667,28 @@
 %   \begin{center}
 %   \begin{tabular}{ll}
 %     \toprule
-%     Entry & Type \\
+%     Entry & Lua type \\
 %     \midrule
-%       \var{config}        & Table   \\
-%       \var{date}          & String  \\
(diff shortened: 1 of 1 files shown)


l3build.dtx
```diff
@@ -1662,32 +1662,35 @@
...
+%       \var{halt-on-error} & |boolean| \\
+%       \var{help}          & |boolean| \\
+%       \var{message}       & |string|  \\
+%       \var{names}         & |table|   \\
+%       \var{quiet}         & |boolean| \\
+%       \var{rerun}         & |boolean| \\
+%       \var{shuffle}       & |boolean| \\
+%       \var{stdengine}     & |boolean| \\
+%       \var{target}        & |string|  \\
+%       \var{texmfhome}     & |string|  \\
```
josephwright, line 1693: Probably we should apply the formatting in the column header using `collcell` - but I guess for the present we can defer that.

PR #370, opened 2024-06-28
Automatic module detection only finds modules

jlaurens:
`listmodules` is turned local.
It looks for any `build.lua` in subdirectories at depth 1.

### DIFF
--- a/l3build-stdmain.lua
+++ b/l3build-stdmain.lua
@@ -28,17 +28,15 @@ local exit   = os.exit
 local insert = table.insert
 
 -- List all modules
-function listmodules()
+local function listmodules()
   local modules = { }
   local exclmodules = exclmodules or { }
-  for entry in lfs.dir(".") do
-    if entry ~= "." and entry ~= ".." then
-      local attr = lfs.attributes(entry)
-      assert(type(attr) == "table")
-      if attr.mode == "directory" then
-        if not exclmodules[entry] then
-          insert(modules, entry)
-        end
+  for entry in tree(".", "*/build.lua") do -- TODO: support "**/..."
+    local module = dirname(entry.cwd)
+    if #module > 2 then -- this is not the current directory
+      module = module:sub(3) -- remove the leading "./"
+      if not exclmodules[module] then
+        insert(modules, module)
       end
     end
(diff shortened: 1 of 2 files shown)


l3build-stdmain.lua
```diff
@@ -28,17 +28,15 @@ local exit   = os.exit
 local insert = table.insert
 
 -- List all modules
-function listmodules()
+local function listmodules()
```
josephwright, line 31: I think this needs to be available generally - one might want it in custom `build.lua` code

l3build-stdmain.lua
```diff
@@ -28,17 +28,15 @@ local exit   = os.exit
...
-      if attr.mode == "directory" then
-        if not exclmodules[entry] then
-          insert(modules, entry)
-        end
+  for entry in tree(".", "*/build.lua") do -- TODO: support "**/..."
+    local module = dirname(entry.cwd)
+    if #module > 2 then -- this is not the current directory
+      module = module:sub(3) -- remove the leading "./"
+      if not exclmodules[module] then
+        insert(modules, module)
```
josephwright, line 39: I don't follow - aren't you risking having unpredictable nesting levels?

l3build-stdmain.lua
jlaurens: I made it local because it was not documented in `l3build.dtx`.
Is it already used? by LaTeX?
As global function the name is not appropriate, imagine the documentation
`listmodules(): automatically detects module`
instead of
`detectedmodules(): list of automatically detected module`

Decision required.

l3build-stdmain.lua
jlaurens: No, "*/..." only looks at the first level contrary to "**/..." that looks at arbitrary depth

josephwright: Ah, right, yes. 

So wider question is why not just add to `exclmodules` as in my suggestion - minimal change and all that.

l3build-stdmain.lua
josephwright: What's wrong with `listmodule()` - it returns a list of modules.

PR #372, opened 2024-06-28
ctanfiles local

jlaurens:
global scope pollution

### DIFF
--- a/l3build-manifest.lua
+++ b/l3build-manifest.lua
@@ -33,6 +33,9 @@ for those people who are interested.
       `l3build-manifest-setup.lua`.
 --]]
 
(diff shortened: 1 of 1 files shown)


l3build-manifest.lua
```diff
@@ -33,6 +33,9 @@ for those people who are interested.
       `l3build-manifest-setup.lua`.
 --]]
 
+local ctanfiles
+local tdsfiles
```
josephwright, line 37: Shouldn't this be local insdie the function, so e.g. `local ctanfiles = {}`?

jlaurens: No because it is shared by different functions of this module

PR #373, opened 2024-06-30
dep_install→unpack_deps

jlaurens:
Having a function that claims to install something whereas it use the unpack target instead of the existing install target is highly misleading

### DIFF
--- a/l3build-aux.lua
+++ b/l3build-aux.lua
@@ -153,21 +153,39 @@ function call(modules, target, opts)
 end
 
 ---Unpack the given dependencies.
----A dependency is the path of a directory relative to the main one.
----@param deps table regular array of dependencies.
+---A dependency is the path of another l3build package
+---relative to the main one, or absolute?
+---@param deps table ordered array of dependencies.
 ---@return number 0 on a successful completion, a non 0 error code otherwise.
 ---@see stdmain, check, unpack, typesetting
 ---@usage Private?
-function dep_install(deps)
+--TODO: Remove this function from the global scope
+function unpack_deps(deps)
   local error_level
   for _, dep in ipairs(deps) do
-    print("Installing dependency: " .. dep)
+    print("Unpacking dependency: " .. dep)
     error_level = run(dep, "texlua " .. get_script_name() .. " unpack -q")
     if error_level ~= 0 then
       return error_level
     end
   end
   return 0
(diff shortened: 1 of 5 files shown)


josephwright:
Not sure I see this as 'highly-misleading' (we are installing in the local build tree), but I also am not tied to the current name. Here, I think you could just make `local` - this is likely one we didn't look at when things were tightened up a bit.

jlaurens:
Changing the name helps in understanding the code
- "install" = install globally
- "unpack" = install locally

This function cannot be made local because it is used in different modules. It can be hidden from the global level but l3build code seems not yet completely ready for that.

FrankMittelbach:
Am 30.06.24 um 12:21 schrieb LAURENS Jérôme:
> Changing the name helps in understanding the code
> 
>   * "install" = install globally
>   * "unpack" = install locally

maybe to you, but actually not to me.

With the current names you have to understand that "unpack" refers to 
"unpacking within l3build"  (and I hope that is documented) and with 
your names you have to understand that "install globally" really means 
"install into your local texmf tree" which is in my eyes 
counterintuitive. Of course, that could be documented to do just that, 
but it doesn't make the name better, imho.

And there are cases when one wants to run just "unpack", e.g., when to 
check that the unpacking generates the right files.

u-fischer:
I do not care much about the name of local, internal functions. But I do find it rather dangerous to change names of functions which are (for whatever reason) visible on the global level for quite some time as you can't know if not someone uses them. As an example: the PR contains a todo: `-- TODO: Hide bundleunpack from the global level`
But the LaTeX base build.lua actually redefines the function and uses a custom `bundleunpack`, and so any change here must be checked against existing build scripts.

jlaurens:
@u-fischer 
Absolutely.
This TODO not should have been removed.

u-fischer:
@jlaurens 
> Absolutely.

Hm. But you seem not have grapped the full implication of my remark. If it is unsafe to change the global `bundleunpack` then it is also unsafe to change the global `dep_install`. For example a short search found this: `latex3/latex3/l3kernel/build.lua:  dep_install(checkdeps)`.

jlaurens:
@u-fischer 
Lines 175+ of `l3build-aux.lua` 

A supplemental question is whether to document if officially or not.

PR #379, opened 2024-07-08
Custom option

jlaurens:
Actually, custom options are documented but not supported.
This PR fixes this problem.

New options must be declared prior to parsing the arguments, it is now done in a new preparation script name `build-pre.lua` near `build.lua` eventually loaded by `l3build-arguments.lua`.

New global function `declare_option(⟨name ⟩,⟨table ⟩)` is used to declare a new option whereas update_option(⟨name ⟩,⟨other table ⟩,⟨provide ⟩) is used to update option data. Both validate their arguments.

Testing is not very well designed and should be integrated in the check target process, a forthcoming feature.

### DIFF
--- a/build-pre.lua
+++ b/build-pre.lua
@@ -1,0 +1,153 @@
+-- Configuration script for LaTeX "l3build" files
+
+do
+  -- this block is testing declare_option
+  -- we temporarily replace option_list to make tests
+  -- TODO: make it a separate test available with target `check``
+  -- using custom test types (forthcoming)
+  local function expect(torf,f, msg)
+    local old_option_list = option_list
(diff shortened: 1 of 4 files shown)


josephwright:
This all looks rather baroque: I am wondering if the best plan is simply to rule out custom options, which don't really work with the structure as it is now (they might have done in the older approach).

FrankMittelbach:
agreed, this is a leftover approach and should be dropped from the documentation rather than implemented.

josephwright:
Ok, so the consensus is to drop the idea from the docs. I’ll leave the PR open until I can do that.

PR #387, opened 2024-10-15
Add --dev switch

josephwright:


muzimuzhi:
Need to also add `latex-dev` to where all supported `checkformat` values are listed:

https://github.com/latex3/l3build/blob/3457c13ccf94588237dd7f501d82f31f1af6b990/l3build.dtx#L434-L436

josephwright:
@muzimuzhi Done

davidcarlisle (approved):

muzimuzhi (approved):

FrankMittelbach:
I think this functionality warrants a ltnews entry (in a different repo I know) explaining how a maintainer of a package using l3build for it can easily test his or her package with the upcoming LaTeX release this way.

l3build.lua
```diff
@@ -131,6 +131,10 @@ if forcedocepoch then
     typesetopts = typesetopts .. " -utc"
   end
 end
+-- Allow for LaTeX "dev" release
+if options["dev"] and checkformat == "latex" then
+  checkformat = "latex-dev"
```
FrankMittelbach, line 136: not sure if it helps or if it makes things worse, but shouldn't there perhaps a terminal output stating --dev not applied if checkformat is not latex?

FrankMittelbach (approved):

l3build.lua
josephwright: I've added one - it only shows at the top level due to the way configs work, but hopefully that's enough?

PR #396, opened 2024-11-18
Correct defaults of `forcecheckepoch` and `forcedocepoch` in doc

muzimuzhi:
They are boolean variables but were documented with string defaults.


muzimuzhi:
Ping.

All other boolean default values are documented without quotes, and quoted values look like string type.
image

PR #417, opened 2025-05-24
Support `exhibitfiles` to set files that need to be "exhibited" in the current workspace; fix a typo

myhsia:
## Support `exhibitfiles` to set files that need to be "exhibited" in the current workspace

Some users merge a `README.md` file, for example, into a `*.dtx` file.
Once they change the markdown file, they should modify the source `*.dtx` file.

But `l3build` won't move the generated (unpacked) markdown file to the main directory (corresponding to `currentdir` in `l3build-variables`), and users usually need to keep the markdown file (or other files) in the main directory so that it can be displayed on the GitHub repository homepage.

So, a variable `exhibitfiles` is added: after `unpack` is executed, the enumerated text files will be copied from `unpackdir` to `currentdir`; and certainly, these `exhibitfiles` will be cleaned after `l3build clean`.

## Fix a typo

In `l3build.dtx`, line 2249: `\loggingoutput` is misspelled in the `macro` environment.


l3build.dtx
```diff
@@ -2651,4 +2655,4 @@
 %
 % \end{implementation}
 %
-% \PrintIndex
+% ^^A \PrintIndex
```
cfr42, line 2658: Removing the index doesn't seem helpful.

l3build-unpack.lua
```diff
@@ -39,6 +39,9 @@ function unpack(sources, sourcedirs)
       return errorlevel
     end
   end
+  for _,i in ipairs(exhibitfiles) do
+    cp(i, unpackdir, currentdir)
```
cfr42, line 43: doesn't this assume `currentdir` is `maindir`, which may often not be the case?

cfr42:
I think this would be better accommodated by implementing existing requests for additional hooks e.g.  `unpackinit_hook()`.

cfr42:
but you need to ask the maintainers to consider this rather than me ;).

josephwright:
I'm closing here: whilst in the past I've used a 'README in `.dtx`' approach, it really isn't a good plan.

PR #422, opened 2025-06-25
Fix CI

muzimuzhi:
- Install `amsfonts` package for `amssymb.sty` needed by building documentation
- Trick the new ConTeXt logging format

In the recent ConTeXt, `\TYPE{OMIT}` would log `just a message  > OMIT`, thus the detection for `^OMIT$` now failed.

I tricked a bit to let the CI pass, but in the long-term if the `just a message  > ` logging prefix is stable and persists, we may need to update the log line cleaning and filtering logic in `l3build-check.lua`.

### DIFF
--- a/.github/tl_packages
+++ b/.github/tl_packages
@@ -15,6 +15,7 @@ etex-pkg
 # Support for typesetting the docs
 #
 alphalph
+amsfonts
 amsmath
 booktabs
 ec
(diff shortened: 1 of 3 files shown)


muzimuzhi:
Resolved in https://github.com/latex3/l3build/compare/a2e7643839ef6a11c1c8ed927931bfb5b7f4314b...92be59778c8ea6da5c24995979be9576f23cb64d, which added support for the updated ConTeXt log formatting. Thanks @josephwright .

PR #425, opened 2025-07-10
Remove the Multi flag from repository

gucci-on-fleek:
CTAN complained when I submitted an update with multiple repository values, so this commit removes the `Multi` flag from the `repository` entry in Table 2 of the l3build manual.

### DIFF
--- a/l3build.dtx
+++ b/l3build.dtx
@@ -1640,7 +1640,7 @@
 %  \texttt{development } &      & \YES & URL(s) of development channels         \\
 %  \texttt{home        } &      & \YES & URL(s) of home page                       \\
 %  \texttt{note        } &      &      & Internal note to CTAN                  \\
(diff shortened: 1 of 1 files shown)


davidcarlisle:
The status of the fields match

https://ctan.org/submit/fields

which lists repository as `list` which corresponds to the `multi` flag.

So I think we should push back to ctan that this should be allowed, or they should change their validator

davidcarlisle:
@FrankMittelbach  we are discussing it with ctan now.
the ctan internal tools are currently not consistent both on whether to allow multiple URL and also if multiple URL are allowed whether that should be a single field with URL separated by `;` (which would allow multiple URL to be specified using the web form https://ctan.org/upload) or whether (as in the submission that triggered this) the  http POST submission has multiple `repository` fields (which can only happen using the http POST API) via curl or curl wrappers such as l3build)

PR #438, opened 2025-12-07
Adjust epoch normalization to include local timezone shift

### LINKED ISSUE #437: Setting the `epoch` value to an ISO-formatted date does not work in timezones ahead of UTC
The manual says (in section 2.8 "Epoch setting"): "The `epoch` variable may be given as a raw value (a simple integer) or as a date in ISO format.". While setting the value to an integer seems to work always, setting the value to an ISO date does not work in certain timezones, it seems. For me, the error `time result cannot be represented in this installation` is raised.

I found that running `\directlua{ tex.print(os.time({year = 1970, month = 1, day = 1, hour = 0, sec = 0, isdst = nil })) }` on my machine results in the error `time result cannot be represented in this installation` (which means that the date is out of range) while setting `hour = 1` returns `0`. This is obviously due to the fact that I am in UTC+1 and `os.time` is timezone aware.

This means that in timezones ahead of UTC, you can't set an ISO date as `epoch`, because this will always call `os_time({ year = 1970, month = 1, day = 1, hour = 0, sec = 0, isdst = nil })` being part of the [function `normalize_epoch(epoch)`](https://github.com/latex3/l3build/blob/92ca3651d706424c64c9d16f98706ccb061371b2/l3build-aux.lua#L47) and thus raise the said error.

Apart from that, it is not directly clear from the short sentence in the manual how exactly a "date in ISO format" should be given  (at least in my opinion). A timestamp in ISO format (such as `2025-12-01T00:00:00Z`  or `20251201T000000Z`), for example, does not work. Only a date formatted as `2025-12-01` would work (but only in certain timezones). An example on how to set the `epoch` value to a ISO-formatted date would make this much clearer.

jasperhabicht:
This is a suggestion to fix #437.

### DIFF
--- a/l3build-aux.lua
+++ b/l3build-aux.lua
@@ -52,12 +52,9 @@ function normalize_epoch(epoch)
   local y, m, d = match(epoch, "^(%d%d%d%d)-(%d%d)-(%d%d)$")
   if y then
     return os_time({
-        year = y, month = m, day   = d,
-        hour = 0, sec = u, isdst = nil
-      }) - os_time({
-        year = 1970, month = 1, day = 1,
-        hour = 0, sec = u, isdst = nil
-      })
+      year = y, month = m, day = d,
+      hour = 0, sec = u, isdst = nil
+    })
   elseif match(epoch, "^%d+$") then
     return tonumber(epoch)
   else


l3build-aux.lua
```diff
@@ -47,15 +48,13 @@ local os_type = os.type
...
   if y then
     return os_time({
-        year = y, month = m, day   = d,
-        hour = 0, sec = 0, isdst = nil
-      }) - os_time({
-        year = 1970, month = 1, day = 1,
-        hour = 0, sec = 0, isdst = nil
-      })
+      year = y, month = m, day = d,
+      hour = 0, sec = u, isdst = nil
```
jasperhabicht: Can we omit `isdst`?

jasperhabicht:
It seems that the underlying issue (#437) is in fact only relevant to Win 10 in certain set-ups. This PR does probably more harm than good and should just be closed, at least for the time being until the real issue has been found.

PR #460, opened 2026-03-30
Versioning var

cfr42:
- `checkpatterns` -> `versionpatterns`
- initialise in `l3build-variables.lua`
- use a differently named local variable in `l3build-check.lua` to combine patterns with those for latex2. latex3 files

I know this is not the model for most variables, but I'd argue the case for treating the version patterns for core LaTeX files by default, even if a user needs to add additional patterns, is strong. That is, I'd much prefer to have the contents of `versionpatterns` added to those defaults, keeping the default value of `versionpatterns` empty. Making users duplicate the patterns for LaTeX2/3 files if they need to add an addition to the list doesn't strike me as sensible,

### DIFF
--- a/l3build-check.lua
+++ b/l3build-check.lua
@@ -123,11 +123,11 @@ local function normalize_log(content,engine,errlevels)
     return false
   end
   -- version/date patterns
-  local versionpatterns = versionpatterns or {}
+  local checkpatterns = checkpatterns or {}
   -- Semantic version-like ones
-  table.insert(versionpatterns, "v%d+%.%d+%.%d+[%d%a.+%-]*")
+  table.insert(checkpatterns, "v%d+%.%d+%.%d+[%d%a.+%-]*")
   -- Classical LaTeX version strings
-  table.insert(versionpatterns, "v%d+%.?%d?%d?%w?")
+  table.insert(checkpatterns, "v%d+%.?%d?%d?%w?")
     -- Substitutions to remove some non-useful changes
   local function normalize(line,lastline,drop_fd)
     if drop_fd then
@@ -181,7 +181,7 @@ local function normalize_log(content,engine,errlevels)
     -- Deal with dates
     if match(line, "[^<]%d%d%d%d[/%-]%d%d[/%-]%d%d") then
       line = gsub(line,"%d%d%d%d[/%-]%d%d[/%-]%d%d","....-..-..")
-      for _,vpattern in ipairs(versionpatterns) do
+      for _,vpattern in ipairs(checkpatterns) do
         line = gsub(line,vpattern,"v...")
       end
     end
--- a/l3build-check.lua
+++ b/l3build-check.lua
@@ -122,6 +122,12 @@ local function normalize_log(content,engine,errlevels)
     end
     return false
   end
+  -- version/date patterns
+  local versionpatterns = versionpatterns or {}
+  -- Semantic version-like ones
+  table.insert(versionpatterns, "v%d+%.%d+%.%d+[%d%a.+%-]*")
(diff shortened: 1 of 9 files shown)


muzimuzhi:
Mention `versionpatterns` in sec. 1.12 "Output normalization", package doc?

Something like

```tex
%   \item Conversion of file dates to \texttt{....-..-..}, and any version
%     numbers on the same lines to \texttt{v...}.
%     If specified, extra version number patterns in \var{versionpatterns}
%     are applied before builtin patterns.
```

https://github.com/latex3/l3build/blob/981385e24be7980b96ff2b4c02cf8c2790a1b7f1/l3build.dtx#L897-L898

l3build-check.lua
```diff
@@ -123,11 +123,11 @@ local function normalize_log(content,engine,errlevels)
...
   end
   -- version/date patterns
-  local checkpatterns = checkpatterns or {}
+  local verpatterns = versionpatterns or {}
   -- Semantic version-like ones
-  table.insert(checkpatterns, "v%d+%.%d+%.%d+[%d%a.+%-]*")
+  table.insert(verpatterns, "v%d+%.%d+%.%d+[%d%a.+%-]*")
   -- Classical LaTeX version strings
-  table.insert(checkpatterns, "v%d+%.?%d?%d?%w?")
+  table.insert(verpatterns, "v%d+%.?%d?%d?%w?")
```
muzimuzhi, line 130: This is done inside the function `normalize_log`. Could be moved to an earlier place so `verpatterns` is only initialized once, at the cost of being global.

For example, in `l3build.lua`, after `l3build-variables.lua` is loaded. (If we want to keep variables initialized in `l3build-variables.lua` mostly public.)

https://github.com/latex3/l3build/blob/981385e24be7980b96ff2b4c02cf8c2790a1b7f1/l3build.lua#L88-L90

l3build.dtx
```diff
@@ -147,7 +147,7 @@
 \luavarset{asciiengines}{\{"pdftex"\}}{Engines which should log as pure ASCII}
 \luavarset{checkruns}   {1}           {Number of runs to complete for a test before comparing the log}
 \luavarset{forcecheckruns}{false}     {Always run \texttt{checkruns} runs and never stop early}
-\luavarset{checkpatterns}{\{ \}}      {Additional file version patterns to normalize when comparing logs}
+\luavarset{versionpatterns}{\{ \}}      {Additional file version patterns to normalize when comparing logs}
```
muzimuzhi:
```suggestion
\luavarset{versionpatterns}{\{ \}}    {Additional file version patterns to normalize when comparing logs}
```

muzimuzhi (approved):

l3build-check.lua
cfr42: I don't see any good reason to put it in `l3build.lua`. As a general principle, I think variables should have the smallest scopes they need. I also think it makes the code easier to read if variables relevant to only one file are defined there (for non-public variables, obviously). 

Originally, I was going to just insert the standard patterns into `versionpatterns`, but I thought that might be a bad idea in case somebody wanted to use different custom patterns for different `check` configs. I don't know if that is likely or not, so maybe accommodating that possibility isn't really necessary? 
- If that isn't desirable, I'd be inclined to get rid of `verpatterns` altogether.
- If it is desirable, then the manipulation should be kept local, though it might well make sense to initialise `verpatterns` with other local initialisations within the file.

What do you think? My instinct is always for maximum flexibility, but I am aware that that is not always the best choice.

muzimuzhi: You are right. I overthought here.

PR #465, opened 2026-06-25
normalize \XeTeXcharclass, Lua-only attribute allocations and user de…

Udi-Fogiel:
…fined whtsit nodes (id and value)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ this project uses date-based 'snapshot' version identifiers.
 ### Changed
 
 - Expand default value of `binaryfiles`
+- Normalize Lua-only attribute, `\XeTeXcharclass`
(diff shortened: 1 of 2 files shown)


u-fischer:
What does that do? How many tlg will it affect?

josephwright:
> What does that do? How many tlg will it affect?

It's more normlisation of numerical data types: as usual, this will show up but is quite predictable.

u-fischer:
Is there an example where it shows up and which explains why it is needed to normalize that? luatex- and xetex-tlgs are normally different anyway.

josephwright:
> Is there an example where it shows up and which explains why it is needed to normalize that? luatex- and xetex-tlgs are normally different anyway.

These are mainly things I've noticed before then forgotten to fix: we do normalise all register numbers, etc., even in engine-specific tests, as that avoids them being affected by other uses of the same data structures.

PR #469, opened 2026-07-12
Re-validate first-time CTAN upload as a new package

### LINKED ISSUE #468: First-time CTAN upload runs the curl config as a shell script and uploads nothing (l3build-upload.lua:185)
When `l3build upload` submits a package that is **not yet on CTAN**, nothing is actually uploaded, yet the target reports success. The failure is in the "re-validate as new package" branch of `upload()` in `l3build-upload.lua:181-186`:

```lua
if override_update_check then
  if match(fp_return,"non%-existent%spackage") then
    print("Package not found on CTAN; re-validating as new package:")
    uploadconfig.update = false
    ctan_post = construct_ctan_post(uploadfile)
    fp_return = shell(ctan_post .. "validate")
  end
end
```

`construct_ctan_post()` returns the curl *config-file content* — a multi-line blob of `form-string="field=value"` lines — not a runnable command. On the first pass, `upload()` writes that blob to the `.curlopt` file and then resets `ctan_post` to `curlexe .. " --config " .. curloptfile .. " https://ctan.org/submit/"` (`l3build-upload.lua:140-164`). This re-validation branch does neither: it neither rewrites the config file nor rebuilds the curl command. So `shell(ctan_post .. "validate")` hands the raw `form-string=...` blob to `/bin/sh -c`, which runs each line as a command:

```
sh: 2: form-string=author=Yegor Bugayenko: not found
sh: 5: form-string=description=This LaTeX package...: File name too long
```

Since `shell()` uses `popen(s,'r')` and captures only stdout, these stderr errors are invisible: `fp_return` is empty, no `ERROR` matches, `exit_status` stays `0`, and the subsequent upload repost at `l3build-upload.lua:222` (`shell(ctan_post .. "upload")`) fails the same silent way. The result is a green build that uploaded nothing. This affects only first-time uploads — existing packages pass the first validation and never enter this branch. A real occurrence is in this build log: https://www.rultor.com/t/59781-4933573080

Suggested fix: in the re-validation branch, rebuild `ctan_post` exactly as the first pass does — write the fresh `construct_ctan_post()` output to `curloptfile` and set `ctan_post = curlexe .. " --config " .. curloptfile .. " https://ctan.org/submit/"` before calling `shell()`. Extracting that write-config-and-assemble-command step into a small helper reused by both call sites would keep them from drifting apart again.

yegor256:
When you upload a package to CTAN for the first time, the first validation pass comes back saying the package does not exist yet. l3build then flips `update` to false and validates again as a new package. That second pass was broken: it called `construct_ctan_post`, which only returns the raw curl config (the `form-string="..."` lines), and handed that straight to `shell`. So instead of running curl, the shell tried to execute each config line as a command, every one failed on stderr, and because `shell` only reads stdout the whole thing looked like it succeeded. The upload reported success and sent nothing.

The initial pass got this right because it wrote the config to a file and built a real `curl --config <file> <url>` command first. This pulls that file-writing and command-assembly step out into `construct_ctan_command` and uses it on both paths, so the re-validation now runs an actual curl command against the submit URL.

Closes #468

### DIFF
--- a/l3build-upload.lua
+++ b/l3build-upload.lua
@@ -137,21 +137,7 @@ function upload(tagnames)
     end
   end
 
-  ctan_post = construct_ctan_post(uploadfile,options["debug"])
-
-
--- curl file version
-  local curloptfile = uploadconfig.curlopt_file or (ctanzip .. ".curlopt")
-  ---@type file*?
-  local curlopt=assert(open(curloptfile,"w"))
-  ---@cast curlopt file*
-  output(curlopt)
-  write(ctan_post)
-  curlopt:close()
-  curlopt = nil
-
-  ctan_post=curlexe .. " --config " .. curloptfile
-
+  ctan_post = construct_ctan_command(uploadfile,options["debug"])
 
 if options["debug"] then
     ctan_post = ctan_post ..  ' https://httpbin.org/post'
@@ -182,7 +168,7 @@ end
     if match(fp_return,"non%-existent%spackage") then
       print("Package not found on CTAN; re-validating as new package:")
       uploadconfig.update = false
-      ctan_post = construct_ctan_post(uploadfile)
+      ctan_post = construct_ctan_command(uploadfile) ..  ' https://ctan.org/submit/'
       fp_return = shell(ctan_post .. "validate")
     end
   end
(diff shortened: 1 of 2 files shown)


yegor256:
@davidcarlisle please, take a look

davidcarlisle (approved):
looks good to me, thanks

josephwright (approved):

l3build-upload.lua
```diff
@@ -182,7 +168,7 @@ end
     if match(fp_return,"non%-existent%spackage") then
       print("Package not found on CTAN; re-validating as new package:")
       uploadconfig.update = false
-      ctan_post = construct_ctan_post(uploadfile)
+      ctan_post = construct_ctan_command(uploadfile) ..  ' https://ctan.org/submit/'
```
muzimuzhi, line 171: Need to pass debugging status here too?

```diff
-      ctan_post = construct_ctan_command(uploadfile) ..  ' https://ctan.org/submit/'
+      ctan_post = construct_ctan_command(uploadfile, options["debug"]) ..  ' https://ctan.org/submit/'
```

PR #471, opened 2026-07-14
Add opt-in typesetwarnings to catch overfull/underfull boxes

### LINKED ISSUE #470: Overfull/Underfull box reports are swallowed by the doc build (exit 0)
The `typeset` function (`l3build-typesetting.lua:124`) returns only the engine errorlevel from `tex()` (`:98`), which is `0` even when the run reports `Overfull`/`Underfull \hbox`/`\vbox`. So `l3build doc`/`ctan` succeed on documentation that contains box-quality problems, and there is no built-in way to make them fail.

This revisits #248, but for a case its resolution never actually covered. #248 was closed WONTFIX on the grounds that warnings are better handled "at the tex level" (redefining `\GenericWarning`/`\msg_warning:nnnn`). That is true for LaTeX/expl3 warnings, but it is technically impossible for overfull/underfull boxes on pdftex and xetex: those messages are emitted by the engine box packer and expose no macro hook, so no `.sty`/driver code can intercept them (a `\GenericWarning`-based package such as `fail-fast` does not catch them). Only LuaTeX offers a hook via the `hpack_quality`/`vpack_quality` callbacks. For the usual pdflatex doc build the builder is therefore the only layer that can help, by inspecting the log.

Concrete, minimal proposal: an opt-in setting (say `typesetwarnings`, default off) that scans the typeset `.log` after a successful run for `Overfull`/`Underfull` lines and returns a non-zero errorlevel when any are found, ideally with a configurable ignore-pattern list. Authors can already achieve this by overriding the global `typeset` function in `build.lua`, so a lighter acceptable outcome would be to simply document that override as the blessed pattern for engine-level warnings rather than add a feature.

yegor256:
This adds a new `typesetwarnings` variable (default `false`). When it is turned on, the `.log` of each typeset file is scanned once compilation finishes and typesetting is reported as failed if it contains any overfull or underfull box lines.

The motivation is the one from #470: overfull and underfull boxes are announced by the TeX engine rather than by a macro, so on pdftex and xetex they cannot be promoted to errors at the TeX level (LuaTeX is the exception, via its `hpack_quality`/`vpack_quality` callbacks). That leaves the builder as the only place able to catch them. The default stays `false`, so existing builds behave exactly as before unless a package opts in.

As discussed in the issue, the code is small and the switch is opt-in, and the manual notes the limitations, including that the check has no effect with `context`, whose log uses different wording. I verified it locally on a small package: with the flag off a document containing an overfull box still exits 0, and with the flag on the same `doc` run fails. No `.tlg` updates are needed since the default behaviour is unchanged.

Closes #470.


yegor256:
@josephwright this implements what we discussed in #470 — opt-in, small, and the docs call out the limitations. Could you take a look when you have a moment? Happy to adjust anything.

PR #472, opened 2026-07-31
Add new BOXTEST environment

hmenke:


testfiles/00-test-1.luatex.tlg
```diff
@@ -17,7 +17,22 @@ TEST 2: Environment test
...
+\hbox(6.94+0.83)x21.0, direction TLT
+.\hbox(0.0+0.0)x0.0, direction TLT
+.\TU/lmtt/m/n/10 #
+.\TU/lmtt/m/n/10 $
+.\TU/lmtt/m/n/10 %
+.\TU/lmtt/m/n/10 &
+! OK.
+\ENDBOXTEST ...depth =\maxdimen \showbox \TESTBOX 
+                                                  \ENDTEST 
+l. ...\ENDBOXTEST
```
hmenke: Not great that this leaks into the log file. This means after we merge this the implementation will basically be frozen, because an update will break all test files.

josephwright: Taking the trick from `expl3`:
```latex
\csname protected\endcsname\long\def\ENDBOXTEST{%
      \ifhmode\unskip\fi
    \egroup
    \begingroup
      \long\def\use##1{##1}%
      \showboxbreadth=\maxdimen
      \showboxdepth=\maxdimen
      \tracingonline=0 %
      \errorcontextlines=-1 %
      \showbox\use{\TESTBOX}%
      \endgroup
  \ENDTEST
}
```
I think would make this rather cleaner: thoughts?

josephwright:
@FrankMittelbach  Does the naming look OK here? (I will review on the code side)

testfiles/00-test-1.luatex.tlg
hmenke: That's perfect! Maybe we should pull the `\showbox` bit into an extra macro for easier reuse?

josephwright: Sure, can do - in `expl3` this is an auxiliary for that reason

l3build.dtx
```diff
@@ -2526,6 +2558,28 @@
     \TYPE{FEATURE UNAVAILABLE}%
   }
 \fi
+\csname newbox\endcsname\TESTBOX% newbox is \outer in plain
+\csname protected\endcsname\long\def\BEGINBOXTEST#1{%
+  \BEGINTEST{#1}%
+    \setbox\TESTBOX=\hbox\bgroup
```
hmenke, line 2564: I spotted a further limitation. First of all this is hardcoded to `\hbox`, which is probably fine, but in XeTeX and LuaTeX we probably want to put directions for bidi.

josephwright: Normally not - you get the prevailing direction, which is almost always what you want (XeTeX doesn't have bidi boxes per se).

l3build.dtx
```diff
@@ -2526,6 +2558,28 @@
...
+  \ENDTEST
+}
+\csname protected\endcsname\long\def\SHOWBOX#1{%
+  \begingroup
+    \long\def\use##1{##1}%
+    \showboxbreadth=\maxdimen
+    \showboxdepth=\maxdimen
+    \tracingonline=0 %
+    \errorcontextlines=-1 %
+    \showbox\use{#1}%
```
muzimuzhi, line 2580: Is the `\use` necessary here? Local test shown that check still passed if `\use` is removed.

Update: Ok, it's required for standalone uses of `\SHOWBOX`. Then no problem.

FrankMittelbach (approved):

PR #476, opened 2026-08-27
Add normalize_log_hook()

josephwright:


gucci-on-fleek:
Perhaps a more flexible interface would be something like the following?

```lua
--[[global]] function default_normalize_line(line,lastline,drop_fd)
    -- The current definition of "local function normalize(line,lastline,drop_fd)"
end

function normalize_line(line,lastline,drop_fd) -- Users can redefine this
    return default_normalize_line(line,lastline,drop_fd)
end
```

Because I can imagine some use-cases where users might want to run their code _before_ the normalization code, or maybe even want to disable it entirely.

josephwright:
> Perhaps a more flexible interface would be something like the following?
> 
> ```lua
> --[[global]] function default_normalize_line(line,lastline,drop_fd)
>     -- The current definition of "local function normalize(line,lastline,drop_fd)"
> end
> 
> function normalize_line(line,lastline,drop_fd) -- Users can redefine this
>     return default_normalize_line(line,lastline,drop_fd)
> end
> ```

How would you cover LuaTeX here?

> Because I can imagine some use-cases where users might want to run their code _before_ the normalization code, or maybe even want to disable it entirely.

I'm a bit wary of allowing removal of normalization - it's part of the documented behaviour of the script (and it's basically all necessary to give usable results).

josephwright:
BTW, we'd also need to allow for a future modification of PDF normalization, so `log` presumably needs to be somewhere in the name.

gucci-on-fleek:
> How would you cover LuaTeX here?

Another level of indirection maybe?

```lua
local _default_normalize_line
function default_normalize_line(...)
    return _default_normalize_line(...)
end

local function normalize_log(content,engine,errlevels)
    local maxprintline = maxprintline
    [...]
    -- Substitutions to remove some non-useful changes
    _default_normalize_line = function (line,lastline,drop_fd)
        if drop_fd then
        [...]
    end
    [...]
end
```

This is admittedly a little ugly though.

> > Because I can imagine some use-cases where users might want to run their code _before_ the normalization code, or maybe even want to disable it entirely.
> 
> I'm a bit wary of allowing removal of normalization - it's part of the documented behaviour of the script 

It's not like it would be exposing internal implementation details that users couldn't otherwise access, since the log output from TeX is already pretty public. But if users explicitly choose to remove the normalization, then they can hardly complain that their logs aren't normalized :)

> (and it's basically all necessary to give usable results)
 
That's the case currently, but largely because (before now) you could only add normalization rules that had almost zero false positives. I could see you potentially adding broader rules in the future if users are able to easily turn them off.

josephwright:
> > (and it's basically all necessary to give usable results)
> 
> That's the case currently, but largely because (before now) you could only add normalization rules that had almost zero false positives. I could see you potentially adding broader rules in the future if users are able to easily turn them off.

I wonder if for that, it would be more sensible to think about a table of rules, so you could disable one at a time. That's more work but would be more flexible.

cfr42:
@gucci-on-fleek There is already an example showing how to remove normalisation on page 25, though it requires declaring a different test type, but I guess you want something which doesn't require that.

cfr42:
I can't currently figure out how to use this. I tried adding to both `build.lua` and `config-tu.lua` and testing with the branch (at least, I think). But I'm not quite sure how to be certain which `l3build-*.lua` it is using.

josephwright:
I take the point about more flexibility, but I think at the moment I'd rather keep things focussed. I looked at using a table: the complexities of normalisation make that quite awkward and I don't think would help.

PR #482, opened 2026-09-03
Add `\REGRESSIONTESTFILEDATE` to hold l3build version

### LINKED ISSUE #456: Record release version in `regression-test.tex`
Can be either stored in a macro, or just as part of the header comment.

This would help in identifying outdated (static) `regression-test.tex` stored and used by repos, for example `latex3/latex2e` and `latex3/latex3`.

muzimuzhi:
Suggestions on a better macro name is very welcome.

- The current `\REGRESSIONTESTFILEDATE` is a bit long, but the naming clash possibility is largely reduced.
- <s>Maybe the suffix `(FILE)DATE` is better than `VERSION`?</s>
- All-caps vs CamelCase (`\RegressionTestFileDate`)?

Closes #456.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ this project uses date-based 'snapshot' version identifiers.
 
 ### Added
 
-- `\REGRESSIONTESTVERSION`, version of `l3build` (issue \#456)
+- `\REGRESSIONTESTFILEDATE`, version of `l3build` (issue \#456)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ this project uses date-based 'snapshot' version identifiers.
 
 ## [Unreleased]
 
+### Added
+
+- `\REGRESSIONTESTVERSION`, version of `l3build` (issue \#456)
+
 ### Changed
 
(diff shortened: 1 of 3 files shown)


josephwright:
Maybe we need to talk about when we update this: I guess when normalization changes?

muzimuzhi:
I didn't follow. Does this related to (log) normalization?

josephwright:
@muzimuzhi  No, I mean that (say) adding a new hook function doesn't change the `.tlg` result, so updating the version string is likely unhelpful, but changing the normalization or result of a `\MACRO` could, so we then would step the value.

muzimuzhi:
Currently the `update_tag()` function in `build.lua` is updated so the `\REGRESSIONTESTFILEDATE` value is always bumped when l3build version is bumped.

I think this mechanism is ok:
- We don't need to maintain two different version dates
- We don't need to think whether a new release may cause test result changes or not
- It provides a debugging possibility for users to check if the versions of `l3build` script and `regression-test.tex` match.

The only drawback is that the content of `regression-test.tex` will change for each release, and it really is not so meaningful to the `regression-test.tex` itself.

Something like `\RegressionTestFileRevision` (CamelCase used for readability) would help, which could be a positive integer and steps each time when `regression-test.tex` 'really' changes (excluding the `\REGRESSIONTESTFILEDATE`).

